PR #23, opened 2020-02-16
New `Universal.ControlStructures.DisallowAlternativeSyntax` sniff

jrfnl:
New sniff to disallow using the alternative syntax for control structures.

Notes:
* This sniff contains a public `allowWithInlineHTML` property to allow alternative syntax when inline HTML is used within the control structure. In all other cases, the use of the alternative syntax will still be disallowed.
* The sniff has modular error codes to allow for making exceptions based on specific control structures and/or specific control structures in combination with inline HTML.

Refs:
* https://www.php.net/manual/en/control-structures.alternative-syntax.php
* squizlabs/PHP_CodeSniffer#2842 - feature request by @jrchamp

Includes fixer.
Includes unit tests.
Includes documentation.
Includes metrics.

### DIFF
--- a/Universal/Docs/ControlStructures/DisallowAlternativeSyntaxStandard.xml
+++ b/Universal/Docs/ControlStructures/DisallowAlternativeSyntaxStandard.xml
@@ -1,0 +1,32 @@
+<?xml version="1.0"?>
+<documentation title="Disallow Alternative Control Structure Syntax">
+    <standard>
+    <![CDATA[
+    The use of the alternative syntax for control structures is not allowed.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: using curly brace syntax for control structures.">
+        <![CDATA[
+if ($foo) <em>{</em>
+    $var = 1;
+<em>}</em>
+
+while (++$i < 10) <em>{</em>
+    echo $i;
+<em>}</em>
+        ]]>
+        </code>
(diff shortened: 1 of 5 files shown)


nikolowry:
@jrfnl do you know if any package provides something like a `Universal.ControlStructures.OnlyAllowAlternativeSyntax`? It makes templating for HTML read so much better -- I'm surprised the only rules I've seen for this center around disallowing it, I want to enforce it everywhere.

jrfnl:
@nikolowry I'm not aware of any such sniffs, though I suppose a sister-sniff could be created, though I'd suggest to then only enforce alternative syntax when there is inline HTML within the control structure.

Maybe open an issue to discuss this further ?

nikolowry:
@jrfnl will open a new issue first thing tomorrow, about to sign off for the night. 

I'm a big believer in always writing code consistently, regardless of one's preferred style -- I'm not trying to add any "micro context-switching" to my work. But i'm sure others would appreciate only enforcing when it's inline html. 

I guess I'm the lone weirdo that thinks the alternative syntax improves readability (read as "scanability"). MediaWiki's argument against the alternative syntax is that "it prevents many text editors from automatically matching and folding brace" -- I use neovim without code folding configured, but I find it had to believe most modern editors can't accommodate.

jrfnl:
@nikolowry Little reminder to open the above mentioned issue ;-)

PR #55, opened 2020-04-29
New `Universal.OOStructures.AlphabeticExtendsImplements` sniff

### LINKED ISSUE #54: New sniff: alphabetize interfaces following `implements` keyword
Similar to the [AlphabeticallySortedUsesSniff](https://github.com/slevomat/coding-standard/blob/master/SlevomatCodingStandard/Sniffs/Namespaces/AlphabeticallySortedUsesSniff.php) in slevomat/coding-standard, I'd like a sniff that checks whether interface names following the `implements` keyword are alphabetically sorted and, if not, fixes them, when running `phpcbf`.

For example, given the following class:

```php
class Baz implements Foo, Bar
{
}
```

This sniff would result in an error because `Foo` and `Bar` are not in alphabetical order.

Using `phpcbf` on this file would correct the code to:

```php
class Baz implements Bar, Foo
{
}
```

jrfnl:
New sniff to verify that the names used in a class "implements" statement or an interface "extends" statement are listed in alphabetic order.

Notes:
* This sniff contains a public `orderby` property to determine the sort order to use for the statement.
    If all names used are unqualified, the sort order won't make a difference.
    However, if one or more of the names are partially or fully qualified, the chosen sort order will determine how the sorting between unqualified, partially and fully qualified names is handled.
    The sniff supports two sort order options:
    - _'name'_ : sort by the interface name only (default);
    - _'full'_ : sort by the full name as used in the statement (without leading backslash).
    In both cases, the sorting will be done using natural sort, case-insensitive.
* The sniff has modular error codes to allow for selective inclusion/exclusion:
    - `ImplementsWrongOrder` - for "class implements" statements.
    - `ImplementsWrongOrderWithComments` - for "class implements" statements interlaced with comments. These will not be auto-fixed.
    - `ExtendsWrongOrder` - for "interface extends" statements.
    - `ExtendsWrongOrderWithComments` - for "interface extends" statements interlaced with comments. These will not be auto-fixed.
* When fixing, the existing spacing between the names in an `implements`/`extends` statement will not be maintained.
    The fixer will separate each name with a comma and one space.
    If alternative formatting is desired, a sniff which will check and fix the formatting should be added to the ruleset.

Includes fixer.
Includes unit tests.
Includes documentation.
Includes metrics.

Fixes #54

### DIFF
--- a/Universal/Docs/OOStructures/AlphabeticExtendsImplementsStandard.xml
+++ b/Universal/Docs/OOStructures/AlphabeticExtendsImplementsStandard.xml
@@ -1,0 +1,23 @@
+<documentation title="Alphabetic Extends Implements">
+    <standard>
+    <![CDATA[
+    The names used in class "implements" or interface "extends" statements should be listed in alphabetic order.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: Names listed alphabetically.">
+        <![CDATA[
+class Baz implements <em>Bar, Foo</em>
+{
+}
+        ]]>
+        </code>
+        <code title="Invalid: Names not listed alphabetically.">
+        <![CDATA[
+class Baz implements <em>Foo, Bar</em>
+{
+}
+        ]]>
+        </code>
+    </code_comparison>
+</documentation>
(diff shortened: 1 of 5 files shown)


ramsey (approved):
This looks perfect to me. Thanks!

jrfnl:
@ramsey Glad to hear it. Feel free to open issues for other sniff ideas you come up and I'll see what I can do ;-)

PR #95, opened 2021-08-16
New `Universal.Files.SeparateFunctionsFromOO` sniff

jrfnl:
This sniff will enforce that a file should either declare (global/namespaced) functions or declare OO structures, but not both.

Nested function declarations, i.e. functions declared within a function/method will be disregarded for the purposes of this sniff.
The same goes for anonymous classes, closures and arrow functions.

Other notes:
- This sniff has no opinion on side effects. If you want to sniff for those, use the PHPCS native `PSR1.Files.SideEffects` sniff.
- This sniff has no opinion on multiple OO structures being declared in one file.
    If you want to sniff for that, use the PHPCS native `Generic.Files.OneObjectStructurePerFile` sniff.

Includes unit tests.
Includes documentation.
Includes metrics.


jrfnl:
@rebeccahum Would love for you to review the sniff! Please feel free to be as critical as can be and ask any questions you may have.

PR #129, opened 2022-09-23
 New `Universal.DeclareStatements.DeclareStatementsStyle` sniff

dingo-d:
This sniff will check the `declare` statements for a specific coding style.

By default, the style of the `declare` statement is without braces and will be applied for all the directives (`strict_types`, `ticks`, `encoding`).

The sniff can be configured to require or disallow braces (except for the `strict_types`), and selectively select for which directive the selected style should be enforced (`ticks`, `encoding`).

- [x] Includes tests
- [x] Includes documentation
- [x] Includes metrics


Universal/Sniffs/DeclareStatements/DeclareStatementsStyleSniff.php
```diff
@@ -0,0 +1,241 @@
...
+
+namespace PHPCSExtra\Universal\Sniffs\DeclareStatements;
+
+use PHP_CodeSniffer\Files\File;
+use PHP_CodeSniffer\Sniffs\Sniff;
+use PHP_CodeSniffer\Util\Tokens;
+use PHPCSUtils\Utils\TextStrings;
+
+/**
+ * Checks the style of the declare statement
```
jrfnl:
```suggestion
 * Checks the style of the declare statement.
```

Universal/Sniffs/DeclareStatements/DeclareStatementsStyleSniff.php
```diff
@@ -0,0 +1,241 @@
...
+ *
+ * `declare` directive can be written in different styles:
```
jrfnl:
```suggestion
 * `declare` directives can be written in two different modes:
```

Universal/Sniffs/DeclareStatements/DeclareStatementsStyleSniff.php
```diff
@@ -0,0 +1,241 @@
...
+ * 1. Applied to the rest of the file, usually written at the top of a file like `declare(strict_types=1);`
+ * 2. Applied to a limited scope using curly braces (the exception to this rule is the `strict_types` directive)
+ * 3. Applied to a limited scope using alternative control structure syntax
```
jrfnl:
```suggestion
 * 1. Applied to the rest of the file, usually written at the top of a file like `declare(strict_types=1);`.
 * 2. Applied to a limited scope using curly braces (the exception to this rule is the `strict_types` directive).
 * 3. Applied to a limited scope using alternative control structure syntax.
```

Universal/Sniffs/DeclareStatements/DeclareStatementsStyleSniff.php
```diff
@@ -0,0 +1,241 @@
...
+     * to all declare directive.
+     *
+     * The exception is the strict_types directive which cannot be
+     * written using curly braces.
+     *
+     * Possible values: ticks, encoding.
+     *
+     * @var array
+     */
+    public $directiveType = [];
```
jrfnl: What about changing this to something along the lines of:
```php
    /**
     * Options: 'disallow', 'allow' (no preference), 'require'.
     */
    public $encodingBlockMode = 'disallow';

    public $ticksBlockMode = 'allow';
    
    private $strictTypesBlockMode = 'disallow';
```

I believe that would make it more intuitive for people to set the properties in their ruleset (and may help with the tests too).

Universal/Tests/DeclareStatements/DeclareStatementsStyleUnitTest.inc
```diff
@@ -0,0 +1,481 @@
+<?php
```
jrfnl: If you like, feel free to have multiple testcase files.

I can imagine you create one base test case file and then copy it over and have different options at the top of the file to verify the different behaviour based on changed settings.

Universal/Sniffs/DeclareStatements/DeclareStatementsStyleSniff.php
```diff
@@ -0,0 +1,241 @@
...
+ * You can also have multiple directives written inside the `declare` directive.
+ * This sniff will check the style of the `declare` directive - by default applied to all directives,
+ * written without braces.
+ *
+ * You can modify the sniff by changing the `declareStyle` option to require or disallow braces,
+ * and require or allow the style to only one of the two directives (`ticks` and `encoding`).
+ *
+ * @since 1.0.0
+ */
+class DeclareStatementsStyleSniff implements Sniff
```
jrfnl: Let's revisit the name once the sniff has had a further iteration. You now duplicate the category name in the sniff name, which isn't needed.

Maybe we should consider something like:
* `Universal.ControlStructures.DeclareStatementMode` or
* `Universal.DeclareStatements.BlockMode`

Universal/Sniffs/DeclareStatements/DeclareStatementsStyleSniff.php
jrfnl: Curly braces and alternative syntax are both a syntax variation of block mode, so should be regarded the same.

Universal/Sniffs/DeclareStatements/DeclareStatementsStyleSniff.php
```diff
@@ -0,0 +1,241 @@
...
+ * 1. Applied to the rest of the file, usually written at the top of a file like `declare(strict_types=1);`
+ * 2. Applied to a limited scope using curly braces (the exception to this rule is the `strict_types` directive)
+ * 3. Applied to a limited scope using alternative control structure syntax
+ *
+ * You can also have multiple directives written inside the `declare` directive.
+ * This sniff will check the style of the `declare` directive - by default applied to all directives,
+ * written without braces.
+ *
+ * You can modify the sniff by changing the `declareStyle` option to require or disallow braces,
+ * and require or allow the style to only one of the two directives (`ticks` and `encoding`).
```
jrfnl: This isn't clear to me, may need some language tweaks once the sniff is finished.

Universal/Sniffs/DeclareStatements/DeclareStatementsStyleSniff.php
```diff
@@ -0,0 +1,241 @@
...
+    /**
+     * Allowed declare directives.
+     *
+     * @var array
+     */
+    private $allowedDirectives = [
+        'strict_types',
+        'ticks',
+        'encoding',
+    ];
```
jrfnl: If the properties are changed, this is probably no longer needed.

Universal/Sniffs/DeclareStatements/DeclareStatementsStyleSniff.php
```diff
@@ -0,0 +1,241 @@
...
+     * @return void
+     */
+    public function process(File $phpcsFile, $stackPtr)
+    {
+        $tokens = $phpcsFile->getTokens();
+
+        $openParenPtr  = $tokens[$stackPtr]['parenthesis_opener'];
+        $closeParenPtr = $tokens[$stackPtr]['parenthesis_closer'];
+
+        if (isset($openParenPtr, $closeParenPtr) === false) {
```
jrfnl: The sniff will already have thrown an "undefined index" notice before reaching this point if one of the two didn't exist....

Universal/Sniffs/DeclareStatements/DeclareStatementsStyleSniff.php
jrfnl: Actually, maybe this should stay for now.... see comment below.

If it stays, it should probably be transformed to an array like the below, so `isset()` can be used:
```
    private $allowedDirectives = [
        'strict_types' => 'strict_types', // <= Can also be set to `true`.
        'ticks'        => 'ticks',
        'encoding'     => 'encoding',
    ];
```

Universal/Sniffs/DeclareStatements/DeclareStatementsStyleSniff.php
```diff
@@ -0,0 +1,241 @@
...
+            return;
+        }
+
+        $directiveStrings = [];
+        // Get the next string and check if it's an allowed directive.
+        // Find all the directive strings inside the declare statement.
+        for ($i = $openParenPtr; $i <= $closeParenPtr; $i++) {
+            if ($tokens[$i]['code'] === \T_STRING) {
+                $directiveStrings[] = $tokens[$i]['content'];
+            }
```
jrfnl:
```suggestion
            if ($tokens[$i]['code'] === \T_STRING
                && isset($this->allowedDirectives[\strtolower($tokens[$i]['content']))
            ) {
                $directiveStrings[] = $tokens[$i]['content'];
            }
```

(something along the lines of...)

Take note of the `strtolower()` and this may need one or two tests as well.

Universal/Sniffs/DeclareStatements/DeclareStatementsStyleSniff.php
```diff
@@ -0,0 +1,241 @@
...
+                        implode(', ', $this->allowedDirectives),
+                        $directiveString
+                    ),
+                    $stackPtr,
+                    'WrongDeclareDirective'
+                );
+                return;
+            }
+            unset($directiveString);
+        }
```
jrfnl: 1. I'm wondering if this is something which belongs in this sniff....
2. If it is maintained, it will have to be moved up into the `for` loop as otherwise it will never trigger (as non-valid `T_STRING`s are no longer saved in the `$directiveStrings` array.

Universal/Sniffs/DeclareStatements/DeclareStatementsStyleSniff.php
```diff
@@ -0,0 +1,241 @@
...
+
```
jrfnl: You may want to add something along the lines of the below here:
```php
if (empty($directiveStrings)) {
    // No valid directives were found, this is outside the scope of this sniff.
    return;
}
```

Universal/Sniffs/DeclareStatements/DeclareStatementsStyleSniff.php
```diff
@@ -0,0 +1,241 @@
...
+                        'Declare directives can be one of: %1$s. "%2$s" found.',
+                        implode(', ', $this->allowedDirectives),
+                        $directiveString
+                    ),
+                    $stackPtr,
+                    'WrongDeclareDirective'
+                );
+                return;
+            }
+            unset($directiveString);
```
jrfnl: For a next time: the `unset()` in this case is not needed within the `foreach()` as each loop overwrites it anyway. It **is** a good practice (👍🏻 ) to unset _after_ the loop, so after the close curly of the `foreach`.

Universal/Sniffs/DeclareStatements/DeclareStatementsStyleSniff.php
```diff
@@ -0,0 +1,241 @@
...
+        }
+
+        foreach ($directiveStrings as $directiveString) {
+            if (!in_array($directiveString, $this->allowedDirectives, true)) {
+                $phpcsFile->addError(
+                    sprintf(
+                        'Declare directives can be one of: %1$s. "%2$s" found.',
+                        implode(', ', $this->allowedDirectives),
+                        $directiveString
+                    ),
```
jrfnl: Not so relevant now as this code will probably be (re)moved, but `addError()` and co have this type of `printf` functionality already build in, so just pass the replacement variables in a `$data` parameter.

See: https://pear.php.net/package/PHP_CodeSniffer/docs/3.5.2/apidoc/PHP_CodeSniffer/File.html#methodaddError

Universal/Sniffs/DeclareStatements/DeclareStatementsStyleSniff.php
```diff
@@ -0,0 +1,241 @@
...
+                    $stackPtr,
+                    'WrongDeclareDirective'
```
jrfnl: Again, not really relevant anymore, but just something to think over:

```suggestion
                    'InvalidDirective'
```

Universal/Sniffs/DeclareStatements/DeclareStatementsStyleSniff.php
```diff
@@ -0,0 +1,241 @@
...
+        // Curly braces.
+        $hasScopeOpenerCloser = isset($tokens[$stackPtr]['scope_opener']);
+
+        // If strict types is defined using curly brace, throw error.
+        if ($hasScopeOpenerCloser !== false && in_array('strict_types', $directiveStrings, true)) {
+            $phpcsFile->addError(
+                sprintf(
+                    'strict_types declaration must not use block mode. Opening brace found on line %d',
+                    $tokens[$stackPtr]['line']
+                ),
```
jrfnl: See my remark about `addError()` having build-in `sprintf()` functionality above.

Universal/Sniffs/DeclareStatements/DeclareStatementsStyleSniff.php
```diff
@@ -0,0 +1,241 @@
...
+                    $stackPtr,
+                    'WrongDeclareDirective'
+                );
+                return;
+            }
+            unset($directiveString);
+        }
+
+        // Curly braces.
+        $hasScopeOpenerCloser = isset($tokens[$stackPtr]['scope_opener']);
```
jrfnl: Nitpick, can be ignored.

```suggestion
        $usesBlockMode = isset($tokens[$stackPtr]['scope_opener']);
```

Universal/Sniffs/DeclareStatements/DeclareStatementsStyleSniff.php
```diff
@@ -0,0 +1,241 @@
...
+
+        // If strict types is defined using curly brace, throw error.
+        if ($hasScopeOpenerCloser !== false && in_array('strict_types', $directiveStrings, true)) {
```
jrfnl: You may want to consider changing how the `$directiveStrings` array is set, to allow for using the more performant `isset()` here.

If you do though, do keep in mind that the keys in that array then need to be lowercased (as otherwise the `isset` comparison won't work) !

Universal/Sniffs/DeclareStatements/DeclareStatementsStyleSniff.php
```diff
@@ -0,0 +1,241 @@
...
+            $phpcsFile->addError(
+                sprintf(
+                    'strict_types declaration must not use block mode. Opening brace found on line %d',
```
jrfnl: > _"Opening brace found on line %d',"_


Is this phrasing technically correct ?

Also: no need to mention the line number as that is a standard part of the error message in PHPCS.

jrfnl: Could this error be made fixable ? (and some of the others too, I suppose)

Universal/Sniffs/DeclareStatements/DeclareStatementsStyleSniff.php
```diff
@@ -0,0 +1,241 @@
...
+                    $tokens[$stackPtr]['line']
+                ),
+                $stackPtr,
+                'Forbidden'
+            );
+            return;
+        }
+
+        // Fix for the case when the code is between the curly braces for the strict_types.
```
jrfnl: Please improve the comment.

Universal/Sniffs/DeclareStatements/DeclareStatementsStyleSniff.php
```diff
@@ -0,0 +1,241 @@
...
+        $codePtr = $phpcsFile->findNext(Tokens::$emptyTokens, ($closeParenPtr + 1), null, true);
```
jrfnl:
```suggestion
        $nextNonEmpty = $phpcsFile->findNext(Tokens::$emptyTokens, ($closeParenPtr + 1), null, true);
```

Universal/Sniffs/DeclareStatements/DeclareStatementsStyleSniff.php
```diff
@@ -0,0 +1,241 @@
...
+
+        /* If the code pointer is not one of:
+         *    \T_SEMICOLON, \T_CLOSE_TAG or \T_OPEN_CURLY_BRACKET, \T_COLON
+         * throw an error.
+         */
```
jrfnl:
```suggestion
        /*
         * If the code pointer is not one of:
         *    \T_SEMICOLON, \T_CLOSE_TAG or \T_OPEN_CURLY_BRACKET, \T_COLON
         * throw an error.
         */
```

I'm wondering if this comment is actually needed ?

Universal/Sniffs/DeclareStatements/DeclareStatementsStyleSniff.php
```diff
@@ -0,0 +1,241 @@
...
+                'Unexpected code found after opening the declare statement without closing it.',
+                $stackPtr,
+                'UnexpectedCodeFound'
+            );
+            return;
+        }
+
+        if ($this->declareStyle === 'requireBraces' &&
+            $hasScopeOpenerCloser === false &&
+            in_array('strict_types', $directiveStrings, true)
```
jrfnl: Nitpick and not enforced yet, but the convention in this repo is to have the boolean at the start of the line, i.e.:
```suggestion
        if ($this->declareStyle === 'requireBraces'
            && $hasScopeOpenerCloser === false
            && in_array('strict_types', $directiveStrings, true)
```

jrfnl (changes requested):
Reviewed during pair programming session in which the review comments were discussed (which is why they are sometimes a bit shorthand).

Additional remarks/things to think over:
* If there is a `declare` statement with multiple directives and those directive have different requirements, i.e, `strict_types`: disallow and `ticks` set to `require`: how should this be handled by the sniff ?

For potentially adding fixers - helpful link: https://pear.php.net/package/PHP_CodeSniffer/docs/3.5.4/apidoc/PHP_CodeSniffer/Fixer.html

General remark: in the sniffs, everything which gets a docblock ought to get a `@since` tag in this repo (can't be enforced yet as there's no good sniff for it, but will be enforced in the future).
For the tests, just the `@since` at class level is fine.

jrfnl:
Some ideas we discussed about adding metrics:

> metric name "Declare statement modus"

> values:
> * block mode
> * single-line

> alternative style values:
> * implode(', ', $directiveStrings) .  block mode
> * implode(', ', $directiveStrings) .  single line mode

dingo-d:
I've reworked the sniff and tests, but still need to work on the fixer and metrics.

Also, I expect there will be some remarks on the PR, those are definitely welcomed 🙂

jrfnl:
FYI: I'm moving this PR out of the `1.0.0-alpha4` milestone as I'm hoping to release over the next few days.

dingo-d:
@jrfnl I've added some metrics, but I'm unsure if this is correct.

When I run the report for metrics, I get correct percentages per file (all adds up to 100%) but I'm not sure if I'm doing it right, as in one file (3rd test file) I have multiple block directives. Some are ok, some are not, and the 2 are ok, but for some reason they are being counted as separate directives, and the type isn't correctly recognized:

```bash
vendor/bin/phpcs --standard=universal --sniffs=universal.declarestatements.blockmode Universal/Tests/DeclareStatements/BlockModeUnitTest.3.inc -sp --report=info

E 1 / 1 (100%)

PHP CODE SNIFFER INFORMATION REPORT
----------------------------------------------------------------------
Declare directive scope: Block [10/10, 100%]

Declare directive type: encoding [4/4, 100%]

----------------------------------------------------------------------
Time: 12.06 secs; Memory: 6MB
```

When debugging, the value is different (first `encoding` then `ticks`) but only `encoding` is reported.

I wanted to group the types, and they seem to be correctly recognized when they are on their own.

I still need to check if there are any fixable errors. The only fixable case I can think of could be if there is only one directive in the `declare` statement that is using block mode when it shouldn't. And then I need to be careful to correctly replace the closer, add semicolon, etc.


### Edit:

I tried to add the fixer, but I got stuck, as there are tons of whitespace edge cases that can trip me up. I did save my WIP for the fixer in another branch so I may play with it later, but for now I'm not sure it's worth the time spent 🤷🏼‍♂️

PR #172, opened 2022-12-02
✨ New Modernize.FunctionCalls.Dirname sniff

jrfnl:
### Introduce new Modernize standard

Includes updating CI and other dev related files to take the new standard into account.

**_P.S.: By rights, this should be a separate standard in the PHPCompatibility organisation, but that's for later. For now, I just want to make sure this sniff becomes available._**

### ✨ New Modernize.FunctionCalls.Dirname sniff

This new sniff will detect and, when possible, auto-fix two typical code modernizations which can be made related to the `dirname()` function:
1. Since PHP 5.3, calls to `dirname(__FILE__)` can be replaced by `__DIR__`.
    Errorcode: `Modernize.FunctionCalls.Dirname.FileConstant`.
2. Since PHP 7.0, nested function calls to `dirname()` can be changed to use the new `$levels` parameter.
    Errorcode: `Modernize.FunctionCalls.Dirname.Nested`.

Depending on supported PHP versions of an application, the complete sniff can be enabled and/or just one of the error codes could be enabled.

**Regarding the fixers:**

The sniff - and more importantly, the fixer - take fully qualified function calls, named parameters, as well as trailing comma's in function calls into account.

* If a comment is found in any of the code which would be removed by the fixer, the issues will still be flagged, but not be marked as fixable.
* If the value of a pre-existing `$levels` parameter cannot be determined, the issues will still be flagged, but not be marked as fixable.

Also take note that the fixer does not make any presumptions about the code style of the project. It is recommended to run the code style fixers together with or after this fixer, so those can adjust the code style to the style applicable in the project.

Includes fixer.
Includes unit tests.
Includes documentation.

#### Notes/Future scope:
* The sniff is not yet compatible with PHPCS 4.x.
* When PHPCSUtils releases its own abstract for sniffing function calls, this sniff should switch to it.

Refs:
* https://www.php.net/manual/en/function.dirname.php
* PHP 5.3: https://php-legacy-docs.zend.com/manual/php5/en/migration53.global-constants
* PHp 7.0: https://www.php.net/manual/en/migration70.changed-functions.php#migration70.changed-functions.core
    Note: the parameter was originally called `$depth`, but has since been renamed to `$levels`.

### DIFF
--- a/Modernize/Sniffs/FunctionCalls/DirnameSniff.php
+++ b/Modernize/Sniffs/FunctionCalls/DirnameSniff.php
@@ -1,0 +1,339 @@
+<?php
+/**
+ * PHPCSExtra, a collection of sniffs and standards for use with PHP_CodeSniffer.
+ *
+ * @package   PHPCSExtra
+ * @copyright 2020 PHPCSExtra Contributors
+ * @license   https://opensource.org/licenses/LGPL-3.0 LGPL3
+ * @link      https://github.com/PHPCSStandards/PHPCSExtra
+ */
+
+namespace PHPCSExtra\Modernize\Sniffs\FunctionCalls;
+
+use PHP_CodeSniffer\Files\File;
+use PHP_CodeSniffer\Sniffs\Sniff;
+use PHP_CodeSniffer\Util\Tokens;
+use PHPCSUtils\Tokens\Collections;
+use PHPCSUtils\Utils\PassedParameters;
+
+/**
+ * Detect `dirname(__FILE__)` and nested uses of `dirname()`.
+ *
+ * @since 1.0.0
+ */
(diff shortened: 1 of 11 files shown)


Modernize/Sniffs/FunctionCalls/DirnameSniff.php
```diff
@@ -0,0 +1,340 @@
...
+
+        $fix = $phpcsFile->addFixableError($error, $stackPtr, $code);
+        if ($fix === false) {
+            return;
+        }
+
+        /*
+         * Fix the error.
+         */
+
```
dingo-d:
```suggestion
```

Extra whitespace

dingo-d (approved):
Found just one whitespace issue.

PR #187, opened 2022-12-08
Modernize/Dirname: magic constants are case-insensitive

jrfnl:
Includes a few updated unit tests to safeguard the fix.


jrfnl:
Coveralls looks to be under attack. As this PR won't affect code coverage, I'm merging it anyway.

image

PR #206, opened 2023-01-05
Enter PHP script names in Composer configuration

szepeviktor:
Composer searches `bin-dir` first

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,25 +39,25 @@
     },
     "scripts" : {
         "lint": [
-            "@php ./vendor/php-parallel-lint/php-parallel-lint/parallel-lint . --show-deprecated -e php --exclude vendor --exclude .git"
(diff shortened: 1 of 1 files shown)


jrfnl:
There's a reason why the commands are like that....

1. `@php` makes sure that the same PHP version will be used as the version used to invoke Composer and not the default PHP version.
2. When using `@php` with the `bin-dir` scripts, the scripts won't work on Windows.

So, sorry, but no, this change is undesired and ill informed.

szepeviktor:
All right. This may be a non viktor compatible project.

jrfnl:
Good PRs are always welcome, however please keep in mind that things are often the way they are for a reason, so just sending in opinionated PRs without actually understanding why things are the way they are, means the chances of the PR(s) getting rejected are very high.

PR #223, opened 2023-03-28
Use PHP syntax highlighting for .inc files

szepeviktor:
Just that.

Now they have a black-and-white look.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -20,6 +20,12 @@ phpunit-bootstrap.php    export-ignore
 /Universal/Tests/        export-ignore
 
 #
+# Use PHP syntax highlighting for these files
+#
+.inc                     linguist-language=PHP
+.inc.fixed               linguist-language=PHP
+
+#
 # Auto detect text files and perform LF normalization
 # http://davidlaing.com/2012/09/19/customise-your-gitattributes-to-become-a-git-ninja/
 #


jrfnl:
Interesting. I've never seen this setting before. Could you point me to the documentation of it ?

szepeviktor:
Here it is, Juliette.
https://github.com/github/linguist/blob/master/docs/overrides.md#summary

jrfnl:
> Here it is, Juliette. https://github.com/github/linguist/blob/master/docs/overrides.md#summary

Thanks Victor. I've had a look at that document and AFAICS it appears to be a GitHub specific feature, which is not [officially supported](https://git-scm.com/docs/gitattributes). Correct ?

I'm also in two minds about the principle of it for PHPCS standard repos:
* Yes, it is nice to see the test case files being properly highlighted and would make the test code more easily readable.
* No, the test case files are not code to be reviewed in the same way as sniff code or even read in the same way.
    Test case files will in most cases contain nonsense code and often contain (intentional) parse errors to make sure the sniff handles those correctly.
    I'm afraid that highlighting those files as "ordinary" PHP files will create more confusion and potential problems (GitHub marking lines as parse errors) than it's worth.
    
I'm going to have to think this over for a while....

szepeviktor:
Seems like this helps, .inc files are full of colors 🎨
https://github.com/PHPCSStandards/PHPCSExtra/blob/c8ba5b37f9a214c4ee6db88b469cf1d0cee80054/.gitattributes#L33

jrfnl:
> Seems like this helps, .inc files are full of colors

Not sure what you mean by that ?

szepeviktor:
.inc files look like this

image

jrfnl:
.. which is okay, as long as it doesn't start annotating "parse errors" and such.

PR #258, opened 2023-07-19
Fix typos

szepeviktor:
Found some misspellings with `typos`.
Code is changed in tests only.


jrfnl (approved):
@szepeviktor Thanks Viktor!

PR #271, opened 2023-09-15
Add Universal.CodeAnalysis.MixedBooleanOperator

TimWolla:
As previously discussed on Fediverse. I've copied over the logic and tests as-is, just making the necessary adjustments to fit the code style of PHPCSExtra, to make it easy to include the Sniff in CodeSniffer itself in the future. 

-----------

This is a clone of the proposed sniff in squizlabs/PHP_CodeSniffer#3205. Adding it to PHPCSExtra to make it generally available quicker than by including it in core CodeSniffer.


Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
```diff
@@ -0,0 +1,107 @@
+<?php
+/**
+ * PHPCSExtra, a collection of sniffs and standards for use with PHP_CodeSniffer.
+ *
+ * @package   PHPCSExtra
+ * @copyright 2021 WoltLab GmbH, 2023 PHPCSExtra Contributors
```
jrfnl: I'd prefer to keep the `@copyright` tags consistent across the codebase, i.e. `@copyright 2023 PHPCSExtra Contributors`.

(both here as well as in the test file)

Universal/Tests/CodeAnalysis/MixedBooleanOperatorUnitTest.inc
```diff
@@ -0,0 +1,41 @@
+<?php
```
jrfnl: As a general "best practice", the initial test files in this repo all follow a similar pattern:

```php
/*
 * OK.
 */
... various code samples which should not be flagged...

/*
 * Bad/The issue
 */
...various code samples which _should_ be flagged.
```

Example: https://github.com/PHPCSStandards/PHPCSExtra/blob/develop/Universal/Tests/Classes/DisallowFinalClassUnitTest.inc

Using this pattern makes it easier for a reader to see what the sniff is looking for/expects.

When it helps further clarify things, I also tend to be quite liberal with inline comments in the tests (also in my code actually and I'm happy to see you are too).

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
```diff
@@ -0,0 +1,107 @@
...
+        $token  = $tokens[$stackPtr];
+
+        $start = $phpcsFile->findStartOfStatement($stackPtr);
+
+        if ($token['code'] === T_BOOLEAN_AND) {
+            $search = T_BOOLEAN_OR;
+        } elseif ($token['code'] === T_BOOLEAN_OR) {
+            $search = T_BOOLEAN_AND;
+        } else {
+            throw new \LogicException('Unreachable');
```
jrfnl: The `process()` method will only ever be called for `T_BOOLEAN_AND` and `T_BOOLEAN_OR` tokens (based on `register()`), so this `else` is redundant.

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
```diff
@@ -0,0 +1,107 @@
...
+     * @return void
+     */
+    public function process(File $phpcsFile, $stackPtr)
+    {
+        $tokens = $phpcsFile->getTokens();
+        $token  = $tokens[$stackPtr];
+
+        $start = $phpcsFile->findStartOfStatement($stackPtr);
+
+        if ($token['code'] === T_BOOLEAN_AND) {
```
jrfnl: Nitpick: while not enforced via a sniff, global constants and calls to global functions are expected to be fully qualified in this repo.
```suggestion
        if ($token['code'] === \T_BOOLEAN_AND) {
```

(here and in various other places in the sniff code)

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
```diff
@@ -0,0 +1,107 @@
...
+                return;
+            } elseif ($tokens[$previous]['code'] === T_CLOSE_PARENTHESIS) {
+                // We skip the content of nested parens.
+                $stackPtr = ($tokens[$previous]['parenthesis_opener'] - 1);
+            } elseif ($tokens[$previous]['code'] === T_CLOSE_SQUARE_BRACKET) {
+                // We skip the content of nested brackets.
+                $stackPtr = ($tokens[$previous]['bracket_opener'] - 1);
+            } elseif ($tokens[$previous]['code'] === $search) {
+                // We reached a mismatching operator, thus we must report the error.
+                $error = "Mixed '&&' and '||' within an expression without using parentheses.";
```
jrfnl:
```suggestion
                $error = "Mixing '&&' and '||' within an expression without using parentheses to clarify precedence.";
```

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
```diff
@@ -0,0 +1,107 @@
...
+            }
+
+            if ($tokens[$previous]['code'] === T_OPEN_PARENTHESIS
+                || $tokens[$previous]['code'] === T_OPEN_SQUARE_BRACKET
+            ) {
+                // We halt if we reach the opening parens / bracket of the boolean operator.
+                return;
+            } elseif ($tokens[$previous]['code'] === T_CLOSE_PARENTHESIS) {
+                // We skip the content of nested parens.
+                $stackPtr = ($tokens[$previous]['parenthesis_opener'] - 1);
```
jrfnl: You are changing the `$stackPtr` position here and on line 96.

If this has happened, the error thrown on line 99 will no longer point to the correct token, which can be confusing in IDEs as well as in the `code` report.

Examples:
image
image

You can see this for yourself by running: `phpcs -sp ./Universal/Tests/CodeAnalysis/MixedBooleanOperatorUnitTest.inc --standard=Universal --sniffs=Universal.CodeAnalysis.MixedBooleanOperator --report=code,source`

With the above in mind, I think it may be better to use a `for()` loop instead of `while` and then adjusting the `$i` variable on this line and on line 96 + `continue`.

When refactoring to a `for` loop, "empty tokens" can be skipped over at the start using `Tokens::$emptyTokens`, though you may want to skip docblocks in one go for efficiency (yes, people sometimes use docblocks _within_ condition expressions 🤷🏻‍♀️ ).

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        while (true) {
+            $previous = $phpcsFile->findPrevious(
+                [
+                    $search,
+                    T_OPEN_PARENTHESIS,
+                    T_OPEN_SQUARE_BRACKET,
+                    T_CLOSE_PARENTHESIS,
+                    T_CLOSE_SQUARE_BRACKET,
```
jrfnl: While I understand you are trying to skip over nested bracketed token groups, your list of tokens to search for is incomplete.

Probably a good idea to use `PHP_CodeSniffer\Util\Tokens::$bracketTokens` + `PHPCSUtils\Tokens\Collections::shortArrayTokens()` here.

Some example code (will probably need better and more test cases, but this is just to trigger your imagination and yes, I know it's silly code, but people write silly code....):
```php
if (true && [true, callMe(), ${$var}] || true);
```

Also: if we can come up with any conceivable code sample where an attribute would be in an expression containing booleans and/or operators, it would probably be a good idea to also skip over attributes.
(and yes, while the code would be silly, it is possible to do this in PHP, so we probably should skip over attributes as I've seen too much silly code in production by now not to take it into account in sniffs)

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
jrfnl: For live coding/code containing a parse error, the `$tokens[$previous]['parenthesis_opener']` key may not be set. The `(else)if` condition should probably take this into account and there should probably also be a test for this.

When adding the test, I find it generally easiest to have parse error tests in separate test case files (`MixedBooleanOperatorUnitTest.2.inc`), especially if multiple tests with parse errors are needed. Each parse error should then get a separate test case file so the behaviour around it can be tested in isolation.

The test framework allows for this. See for example, [this test file](https://github.com/PHPCSStandards/PHPCSExtra/blob/develop/Universal/Tests/Classes/DisallowAnonClassParenthesesUnitTest.php) on how to handle this.

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
```diff
@@ -0,0 +1,107 @@
...
+            } elseif ($tokens[$previous]['code'] === T_CLOSE_SQUARE_BRACKET) {
+                // We skip the content of nested brackets.
+                $stackPtr = ($tokens[$previous]['bracket_opener'] - 1);
+            } elseif ($tokens[$previous]['code'] === $search) {
+                // We reached a mismatching operator, thus we must report the error.
+                $error = "Mixed '&&' and '||' within an expression without using parentheses.";
+                $phpcsFile->addError($error, $stackPtr, 'MissingParentheses', []);
+                return;
+            } else {
+                throw new \LogicException('Unreachable');
```
jrfnl: Throwing exceptions in PHPCS code is a big no-no. PHPCS will catch the exception, but will then stop the scan of the file completely for the file which caused the exception to be triggered, hiding any errors _after_ the code which caused the exception to be triggered.

Silently ignoring code which cannot be analyzed/parse error code is the preferred way to go.

Universal/Tests/CodeAnalysis/MixedBooleanOperatorUnitTest.inc
```diff
@@ -0,0 +1,41 @@
...
+);
+
+if (true && foo(true || true));
+if (true && foo(true && true || true));
+if (true && $foo[true || true]);
+if (true && $foo[true && true || true]);
+
+if (true && foo(true) || true);
+if (true && $foo[true] || true);
+if (true && foo($foo[true]) || true);
```
jrfnl: Might it be an idea to have some more variation of control structures and other code which could have boolean operators in them here ?

That might help catch issues not currently anticipated.

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
```diff
@@ -0,0 +1,107 @@
...
+     *                                               in the stack passed in $tokens.
+     *
+     * @return void
+     */
+    public function process(File $phpcsFile, $stackPtr)
+    {
+        $tokens = $phpcsFile->getTokens();
+        $token  = $tokens[$stackPtr];
+
+        $start = $phpcsFile->findStartOfStatement($stackPtr);
```
jrfnl:
```suggestion
        $start = BCFile::findStartOfStatement($phpcsFile, $stackPtr);
```

(with `use PHPCSUtils\BackCompat\BCFile;`)

This safeguards that if/when that function would change upstream, the upstream change will be available on older PHPCS versions.

jrfnl (changes requested):
Hi @TimWolla, oh well, figured I may as well review it straight away. Thanks for getting this PR transferred over!

### Naming of the sniff

Question: should the sniff name be `MixedBooleanOperator` or `MixedBooleanOperators` (note the plural) ?
Personally, I'm leaning towards the plural form as if there is only one operator, the issue doesn't exist and there will never be a mix.

Note: we need to settle on a good name now as renaming the sniff later would be a breaking change.

### Premise of the sniff

As things are, the sniff only concerns itself with the boolean and/or operators.
Questions:
* Should it be documented (in the class docblock) why the boolean not `!` operator is not included in this sniff ?
* Should the sniff also apply to the logical `and`/`or`/`xor` operators ?
    If not, maybe we should open an issue to create a sister-sniff for the logical operators ?

### Review notes:

Docs: checked against the typical issues (line length too long, non-space indentation, missing `<em>` tags etc) and everything is 💯 okay.

Sniff & tests: see my inline notes.

Metrics are not relevant for this sniff. ✔️ 
Auto-fixing could be possible for this sniff, but would involve an assumption about the correctness of the code as written, so I'm fine with this sniff not having a fixer and letting a human decide where the parentheses should be. ✔️ 

I'm also trying to think if there is a situation in which the sniff should stop scanning or behave differently when an operator with a higher precedence is encountered ?
Note: the `$phpcsFile->findStartOfStatement($stackPtr)` call does not take this into account, so the sniff would have to.


### Other

The build failures related to the code coverage checks are a known issue for PRs from forks and I'm working with the Coveralls team on a solution. You don't need to worry about this (though checking code coverage locally for the sniff would be good and will show that the sniff as-is is not completely covered by tests).

image


Hope this review helps. Please let me know if you have any questions or would like me to clarify any of my remarks.

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
TimWolla: As indicated by the exception message, I assume the `else` branch to never be reached. If it would be reached, it would indicate a programming error on my part.

In my code I've adopted this pattern to make sure any such programming error becomes immediately apparent, instead of the code silently misbehaving.

That said: I can completely remove the `else` branch if that is what you prefer.

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
TimWolla: See the response for the other `Unreachable` exception. If PHP 8.0 would be the minimum requirement, I would've written this as a `match()` which would likewise Error for an unhandled case.

Universal/Tests/CodeAnalysis/MixedBooleanOperatorUnitTest.inc
TimWolla: The examples are currently grouped to have matching “not ok” and “ok” examples right beneath each other (i.e. first the one without parentheses and then the two possible variants with parentheses).

I can reorder the tests if you insist, but I personally find that harder to understand. It would also shift all line number if tests are added after the fact.

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
TimWolla: I'm afraid, my terminal doesn't show the underlining when running your command, so I'm unable to see this for myself:

image

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
jrfnl: I understand your reasoning, but if you haven't been able to come up with (test) code which would trigger this, this will never be triggered in a dev situation, so the only time the Exception gets thrown is at runtime for end-users, which causes a problem in _their_ code scans/CI as the file under scan doesn't get scanned completely anymore.

You could leave the `else`, but replace the `throw ...` statement with a `return`, which will not cause a problem for end-users.

In those cases, I normally accompany the `return` with a `// Shouldn't be possible, but just in case.` comment + a `// @codeCoverageIgnore`.

Universal/Tests/CodeAnalysis/MixedBooleanOperatorUnitTest.inc
jrfnl: Okay, let's leave them in the order they are then, but in that case, what about having trailing `// Okay.` / `// Error.` comments ? That way it is still clearer what is being tested in each line.

jrfnl: > It would also shift all line number if tests are added after the fact.

It is only for the initial set of tests. Tests added at a later point go after with a comment referencing the issues they relate to.

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
TimWolla: Would the use of `PHPCSUtils\Tokens\Collections::shortArrayTokens()` make it harder to upstream the Sniff? After all, this is my main goal. 

> Some example code (will probably need better and more test cases, but this is just to trigger your imagination and yes, I know it's silly code, but people write silly code....):

Thanks, added this as a test case.

> Also: if we can come up with any conceivable code sample where an attribute would be in an expression containing booleans and/or operators, it would probably be a good idea to also skip over attributes.

Thanks, I hate writing such code, but there is indeed a code sample that allows for attributes. It's already correctly handled.

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
jrfnl: > ... which would likewise Error for an unhandled case.

In which case I would have suggested wrapping it in a `try/catch` if I thought it would be possible for the error to be thrown ;-)

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
TimWolla: Okay, removed.

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
jrfnl: Would you like me to give you a screenshot of the complete output ?

All the same, the issue with the `$stackPtr` being changed should be clear enough, even without seeing it in the terminal ?

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
jrfnl: > Would the use of `PHPCSUtils\Tokens\Collections::shortArrayTokens()` make it harder to upstream the Sniff? After all, this is my main goal.

Not really, it's just shorthand for the tokens used for short arrays. For upstream, you'd need to explicitly add those tokens, but PHPCSExtra can use the shorthand.

https://github.com/PHPCSStandards/PHPCSUtils/blob/77a6aa4e6abc437eb0cb71702c1a635655a33af8/PHPCSUtils/Tokens/Collections.php#L499-L502

> Thanks, I hate writing such code, but there is indeed a code sample that allows for attributes. It's already correctly handled.

Yup... I've learned to write really, really, really horrible code just to test sniffs, the joys of maintaining tooling like this... 😂

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
TimWolla: > Would you like me to give you a screenshot of the complete output ?

image

> All the same, the issue with the $stackPtr being changed should be clear enough, even without seeing it in the terminal ?

Yes, it's clearly described. However it would be useful to see the behavior myself when making the adjustments to verify the adjustments are actually correct or even the best behavior. Currently the code scans *backwards* to find offending boolean operators, which in `$foo = true || false || true && false;` probably reports the second `||` which is right in the middle. It might be better to scan forwards, which would reliably report the `&&` which makes more sense.

Universal/Tests/CodeAnalysis/MixedBooleanOperatorUnitTest.inc
TimWolla: I can add some to make sure, but other control structures are not really necessary to test, because the “working space” of the Sniff is limited to a single expression and the stuff that is legal to put into an expression is clearly delineated by either the semicolon or the various parentheses / brackets.

TimWolla:
Handled the obvious stuff already. Will look into the stuff that needs some additional thinking later. Thank you for the review so far!

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
TimWolla: Narf. Now PHPStan is unhappy. I'll add the `return` then, because the `elseif` IMO improves clarity.

Universal/Tests/CodeAnalysis/MixedBooleanOperatorUnitTest.inc
jrfnl: In which case, I suggest adding a test with something along the lines of the below and to see if your statement still holds up....:
```php
for ( $i = true && false, $j = false || true; $i && $j || $k, $i < $j; $i++ ) {}
```

Oh and some tests with ternaries would also not go amiss....

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
TimWolla: `--colors` fixed it.

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
```diff
@@ -0,0 +1,95 @@
...
+            $stackPtr,
+            $start,
+            false,
+            null,
+            true
+        );
+
+        if (
+            $previous === false
+            || \in_array($tokens[$previous]['code'], [\T_INLINE_THEN, \T_INLINE_ELSE], true)
```
TimWolla: Should these be implicitly handled when setting `local` to `true` for `findPrevious`?

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
```diff
@@ -0,0 +1,95 @@
...
+     */
+    public function register()
+    {
+        return [
+            \T_BOOLEAN_OR,
+            \T_BOOLEAN_AND,
+            \T_LOGICAL_AND,
+            \T_LOGICAL_OR,
+            \T_LOGICAL_XOR,
+        ];
```
jrfnl:
```suggestion
        return Tokens::$booleanOperators;
```

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
```diff
@@ -0,0 +1,95 @@
...
+        $previous = $phpcsFile->findPrevious(
+            \array_filter(
+                \array_merge(
+                    $this->register(),
+                    [\T_INLINE_THEN, \T_INLINE_ELSE]
+                ),
+                function ($token) use ($valid) {
+                    return $token !== $valid;
+                }
+            ),
```
jrfnl: Sniffs should be very conscious about performance as they are called every single time a certain token is encountered.
This also means that this bit of code will be run _every single time_ the tokens are encountered, which seems unnecessary as we're basically talking a largely fixed array.

A typical pattern to improve performance for this code would be as follows (pseudo code to explain):
```php
    private $searchTargets;

    public function register()
    {
        // Set $searchTargets only once.
        $this->searchTargets = Tokens::$booleanOperators;
        $this->searchTargets[\T_INLINE_THEN] = \T_INLINE_THEN;
        $this->searchTargets[\T_INLINE_ELSE] = \T_INLINE_ELSE;

        return Tokens::$booleanOperators;
    }

    public function process()
    {
        ...
        $valid = $token['code'];
        $search = $this->searchTargets;
        unset($search[$valid]);

        $previous = $phpcsFile->findPrevious($search ....);
        ...
    }
```

jrfnl (changes requested):
I haven't done a full re-review yet, but thought I'd give you the below feedback already.

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
TimWolla: Thanks, I could simplify this and improve the output at the same time:

```
if (true || true && true && true && true); // Not OK.
```

previously this reported each `&&` separately, now only the first one is reported. This should result in less now, because the developer will need to look at the expression anyway.

TimWolla:
> ### Naming of the sniff
> 
> Question: should the sniff name be `MixedBooleanOperator` or `MixedBooleanOperators` (note the plural) ? Personally, I'm leaning towards the plural form as if there is only one operator, the issue doesn't exist and there will never be a mix.
> 

This makes sense. Should this also include the “Binary” part within the name? i.e. MixedBinaryBooleanOperators?

> ### Premise of the sniff
> 
> As things are, the sniff only concerns itself with the boolean and/or operators. Questions:
> 
>     * Should it be documented (in the class docblock) why the boolean not `!` operator is not included in this sniff ?

With including the Binary in the name this should be clear. But I can add a short blurb (or you add some suggestion using GitHub's feature.

>     * Should the sniff also apply to the logical `and`/`or`/`xor` operators ?
>       If not, maybe we should open an issue to create a sister-sniff for the logical operators ?

Included now.

> 
> ### Review notes:
> 

> Auto-fixing could be possible for this sniff, but would involve an assumption about the correctness of the code as written, so I'm fine with this sniff not having a fixer and letting a human decide where the parentheses should be. ✔️

Yes, this Sniff is intentionally written for detecting possibly *incorrect* code and we found several instances in our code base with the initial version I proposed. That's why autofixing must not happen.

> I'm also trying to think if there is a situation in which the sniff should stop scanning or behave differently when an operator with a higher precedence is encountered ? Note: the `$phpcsFile->findStartOfStatement($stackPtr)` call does not take this into account, so the sniff would have to.

The Sniff tripping over such a situation would likely indicate that the human would also be confused. I am fine with some false-positives should there be such a situation, you can't have enough parentheses to make precedence clear. And if you mix `and` and `&&` or others, you are really asking for trouble if you don't use parentheses.

> ### Other
> 
> The build failures related to the code coverage checks are a known issue for PRs from forks and I'm working with the Coveralls team on a solution. You don't need to worry about this (though checking code coverage locally for the sniff would be good and will show that the sniff as-is is not completely covered by tests).
> 

I wasn't able to get coverage running locally, due to the ancient PHPUnit requirement that is not trivially compatible with PHP 8.2. But with the updated code there isn't really much logic left in the sniff itself. If you would check that it's up to 100% coverage now, I'd appreciate it.

TimWolla:
The commit history is a little messy due to the number of changes required due to my lack of knowledge of CodeSniffer. I can squash it for you once you are happy with the diff.

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
jrfnl: PHPCS does not take these into account. Whether it should or not, is a different debate and I'd rather not pollute this PR with that discussion as we can't solve it in the PHPCSExtra repo anyway.

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
jrfnl: Generally speaking, hiding one error "behind" another is a no-no for sniffs (as people will fix something and then, surprise!, a new error along the same lines pops up).

Having said that, in this case I think it makes sense to only flag the first operator which is different, so all good.

jrfnl:
> The commit history is a little messy due to the number of changes required due to my lack of knowledge of CodeSniffer. I can squash it for you once you are happy with the diff.

Yes please. Would be great if you could do that once the PR is ready for merge.

Universal/Docs/CodeAnalysis/MixedBooleanOperatorStandard.xml
jrfnl, line 1: Would it make sense to have some variation in the code samples ? They now all show assignments, which could mislead people into thinking the sniff only applies to assignments.

Maybe change one or two code samples to `if() {}` statements ?

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
```diff
@@ -0,0 +1,89 @@
...
+
+use PHP_CodeSniffer\Files\File;
+use PHP_CodeSniffer\Sniffs\Sniff;
+use PHP_CodeSniffer\Util\Tokens;
+use PHPCSUtils\BackCompat\BCFile;
+
+/**
+ * Forbid mixing different binary boolean operators within a single expression without making precedence
+ * clear using parentheses.
+ *
```
jrfnl: To add:
```
Sister-sniff to the PHPCS native `Squiz.ControlStructures.InlineIfDeclaration` and `Squiz.Formatting.OperatorBracket.MissingBrackets` sniffs.
```

FYI: the first is a sniff which expects the condition of a ternary to be within parentheses. The second is a sniff which expects arithmetic expressions to be within parentheses.

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
```diff
@@ -0,0 +1,89 @@
...
+ * @link https://github.com/squizlabs/PHP_CodeSniffer/pull/3205
+ * @link https://github.com/php-fig/per-coding-style/issues/22
+ *
+ * @since 1.2.0
+ */
+final class MixedBooleanOperatorSniff implements Sniff
+{
+    private $previousTokens = [];
```
jrfnl: 1. Please add a docblock for the property.
2. If I'm honest, I think the `$previousTokens` name is confusing as the array does not contain the tokens seen previously, but the search targets for a call to `findPrevious()`.

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
```diff
@@ -0,0 +1,89 @@
...
+     * @since 1.2.0
+     *
+     * @return array<int|string>
+     */
+    public function register()
+    {
+        $this->previousTokens = \array_merge(
+            Tokens::$booleanOperators,
+            [\T_INLINE_THEN, \T_INLINE_ELSE]
+        );
```
jrfnl: Honest question: why are you (still) using `array_merge()` here ?

Not a problem for the current state of the sniff, but it is an anti-pattern in the context of PHPCS and token arrays, so I'm curious to know your reason (and happy to provide an explanation of why it is an anti-pattern too, but I didn't want to infuence your answer by providing that already).

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
```diff
@@ -0,0 +1,89 @@
...
+            true
+        );
+
+        if (
+            // No token found.
+            $previous === false
+            // Identical operator found.
+            || $tokens[$previous]['code'] === $tokens[$stackPtr]['code']
+            // Beginning of the expression found for the ternary conditional operator.
+            || \in_array($tokens[$previous]['code'], [\T_INLINE_THEN, \T_INLINE_ELSE], true)
```
jrfnl: Might just be me, but why use `in_array()` when a plain comparison will do ? It's not as if we need to search a huge array.
```suggestion
            || $tokens[$previous]['code'] === \T_INLINE_THEN
            || $tokens[$previous]['code'] === \T_INLINE_ELSE
```

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
```diff
@@ -0,0 +1,89 @@
...
+        ) {
+            return;
+        }
+
+        // We found a mismatching operator, thus we must report the error.
+        $error = "Mixing different binary boolean operators within an expression without using parentheses to clarify precedence.";
+        $phpcsFile->addError($error, $stackPtr, 'MissingParentheses', []);
```
jrfnl:
```suggestion
        $phpcsFile->addError($error, $stackPtr, 'MissingParentheses');
```

No need to pass an unused optional parameter.

jrfnl (changes requested):
Hi @TimWolla Thank you for making those updates. Feels like we're very close to the final version now.

I've gone through the sniff with a fine toothcomb again. Hope you find my notes helpful.

> > ### Naming of the sniff
> > Question: should the sniff name be `MixedBooleanOperator` or `MixedBooleanOperators` (note the plural) ? Personally, I'm leaning towards the plural form as if there is only one operator, the issue doesn't exist and there will never be a mix.
>
> This makes sense. Should this also include the "Binary" part within the name? i.e. MixedBinaryBooleanOperators?

I'm not sure where the "Binary" terminology comes from ? The PHP manual calls the complete set of these operators ["Logical operators"](https://www.php.net/manual/en/language.operators.logical.php), so I think using that term would prevent confusion.

:point_right: _Note: this will also need to be changed in the error message, README description and the XML docs._


Regarding the name of the sniff, what with the functionality of the sniff having changed, I'm not sure about the name.

The sniff has changed from its original state to now be more comprehensive with the addition of checking for the logical `and`, `or` and `xor` operators.

It is still about mixing operators within a single expression, but more operators are now considered.

So let's turn it around: what is the aim of the sniff ? If I would paraphrase, it is about being explicit about the precedence order of (part of) statements by using parentheses. Do you agree ?
So maybe the name of the sniff should reflect that instead ?

Maybe something along the lines of `RequireExplicitLogicalOperatorPrecedence` ? (Note: I'm definitely not saying that this is the name to go with, I'm just throwing it out here to trigger your imagination)

:reminder_ribbon: _Similar to the terminology change, once the sniff name is settled on, the file names, class names, README and XML docs title will all need to be updated._


> 
> > ### Premise of the sniff
> > As things are, the sniff only concerns itself with the boolean and/or operators. Questions:
> > ```
> > * Should it be documented (in the class docblock) why the boolean not `!` operator is not included in this sniff ?
> > ```
>
> With including the Binary in the name this should be clear. But I can add a short blurb (or you add some suggestion using GitHub's feature.

A short phrase in the class docblock would be good, as the `!` operator is also considered a logical operator.


>
> > ### Review notes:
>
> Yes, this Sniff is intentionally written for detecting possibly _incorrect_ code and we found several instances in our code base with the initial version I proposed. That's why autofixing must not happen.

:+1:

> you can't have enough parentheses to make precedence clear. And if you mix and and && or others, you are really asking for trouble if you don't use parentheses.

Hear, hear! I fully, 100% agree with you on this!

> > I'm also trying to think if there is a situation in which the sniff should stop scanning or behave differently when an operator with a higher precedence is encountered ? Note: the `$phpcsFile->findStartOfStatement($stackPtr)` call does not take this into account, so the sniff would have to.
>
> The Sniff tripping over such a situation would likely indicate that the human would also be confused. I am fine with some false-positives should there be such a situation...

I've just ran the sniff over some large sets of code bases I have locally (including Symfony, Laravel, Drupal, Joomla, WordPress).
The test runs scanned more than 70K files in total and flagged ~600 issues, which I've visually checked (using the `--report=code` view).
I'm now trying hard to forget some of the horrific code I saw... :joy:

I've only found one false positive - see the below code.

```php
return match (true) {
    \defined($value) => \constant($value),
    'yes' === $lowercaseValue,
    'on' === $lowercaseValue => true,
    'no' === $lowercaseValue,
    'off' === $lowercaseValue,
    'none' === $lowercaseValue => false,
    isset($value[1]) && (
        ("'" === $value[0] && "'" === $value[\strlen($value) - 1]) || // <- Here the `||` gets flagged.
        ('"' === $value[0] && '"' === $value[\strlen($value) - 1]) || // <- Here the `&&` gets flagged.
    ) => substr($value, 1, -1), // quoted string
    default => doSomething($value),
};
```

... and while this _could_ also be considered a false positive, I'm inclined to say it should be flagged anyway.
```php
if (
    [] !== $fields &&
    $constraint &&
    $constraint->payload &&
    // If some or all payload fields are whitelisted, add them
    $payloadFields = null === $fields || true === $fields ? $constraint->payload : array_intersect_key($constraint->payload, $fields)
) {}
```

Obviously finding false negatives is not as easy, but the current version of the sniff is looking good based on these tests.

Just thinking, if we can come up with a relevant code sample, it might also be good to add a test with an arrow function ?

Arrow functions are not 100% supported in the tokenizer layer (best effort, but the fact they can share their "scope closer" with other structures make things very difficult), so I think it may be a good idea to have some tests involving arrow functions just to be on the safe side.




> > ### Other
> > The build failures related to the code coverage checks are a known issue for PRs from forks and I'm working with the Coveralls team on a solution. You don't need to worry about this (though checking code coverage locally for the sniff would be good and will show that the sniff as-is is not completely covered by tests).
>
> If you would check that it's up to 100% coverage now, I'd appreciate it. But with the updated code there isn't really much logic left in the sniff itself.

Here's a screenshot of the current state (all good :heavy_checkmark:):

image

> I wasn't able to get coverage running locally, due to the ancient PHPUnit requirement that is not trivially compatible with PHP 8.2.

Yes, I'm aware of this. I'm locally running with the [commits from this upstream PR](https://github.com/squizlabs/PHP_CodeSniffer/pull/3803) included in a local "all my open PHPCS PRs" branch, which means I can run on PHPUnit 9 ;-)

I still need to write up a contributor doc at some point for this repo, but the test setup in this repo allows for running the tests with a local PHPCS repo clone. To do this, you need to copy the `phpunit.xml.dist` file to `phpunit.xml` and then add the following to that file (replacing the path, of course, with the path to the PHPCS clone on your system):
```xml
    <php>
        <env name="PHPCS_DIR" value="path/to/PHP_CodeSniffer/"/>
    </php>
```

That setup allows me to test with whatever branch I check out in my PHPCS clone, which is often useful if a bug in a sniff in an external standard is related to an upstream bug.

TimWolla:
Some preliminary answers, will look into the code stuff later.

> > This makes sense. Should this also include the "Binary" part within the name? i.e. MixedBinaryBooleanOperators?
> 
> I'm not sure where the "Binary" terminology comes from ? The PHP manual calls the complete set of these operators ["Logical operators"](https://www.php.net/manual/en/language.operators.logical.php), so I think using that term would prevent confusion.

It's “Binary Operator” as in “Operator that takes two operands”. `!` would be a unary operator, as it only operates on one operand. And `? :` is a ternary operator, because it operates on three operands. That's why it's called that.

see: https://en.wikipedia.org/wiki/Binary_operation

> Regarding the name of the sniff, what with the functionality of the sniff having changed, I'm not sure about the name.
> 
> The sniff has changed from its original state to now be more comprehensive with the addition of checking for the logical `and`, `or` and `xor` operators.
> 
> It is still about mixing operators within a single expression, but more operators are now considered.
> 
> So let's turn it around: what is the aim of the sniff ? If I would paraphrase, it is about being explicit about the precedence order of (part of) statements by using parentheses. Do you agree ? So maybe the name of the sniff should reflect that instead ?

Yes, that description is accurate. Specifically explicit precedence for binary boolean operators, because for `+` vs `*` the precedence rules are fairly well-known, where-as for `&&` and `||` it's less obvious. `!` is not included because the way it's commonly used indicates that it binds tightly (less tight than instanceof, though. `!$foo instanceof Bar` works correctly in PHP, not in JavaScript though :grin:)

> Maybe something along the lines of `RequireExplicitLogicalOperatorPrecedence` ? (Note: I'm definitely not saying that this is the name to go with, I'm just throwing it out here to trigger your imagination)

Naming, one of the hard things in computer science. I don't particularly care about the name and your suggestion would be fine with me (possibly including the “binary” part). You probably know the naming of existing Sniffs better than me.

> > With including the Binary in the name this should be clear. But I can add a short blurb (or you add some suggestion using GitHub's feature.
> 
> A short phrase in the class docblock would be good, as the `!` operator is also considered a logical operator.

That's where the “Binary” would come in, but I'll add something.

> I've only found one false positive - see the below code.
> 
> ```
> return match (true) {
>     \defined($value) => \constant($value),
>     'yes' === $lowercaseValue,
>     'on' === $lowercaseValue => true,
>     'no' === $lowercaseValue,
>     'off' === $lowercaseValue,
>     'none' === $lowercaseValue => false,
>     isset($value[1]) && (
>         ("'" === $value[0] && "'" === $value[\strlen($value) - 1]) || // <- Here the `||` gets flagged.
>         ('"' === $value[0] && '"' === $value[\strlen($value) - 1]) || // <- Here the `&&` gets flagged.
>     ) => substr($value, 1, -1), // quoted string
>     default => doSomething($value),
> };
> ```

This code is not syntactically valid, as the `||` is directly followed by `) => substr($value, 1, -1), // quoted string`. I did not yet test the sniff with syntactically invalid code. Did you accidentally remove some bits while copying the code into the issue?

> ... and while this _could_ also be considered a false positive, I'm inclined to say it should be flagged anyway.
> 
> ```
> if (
>     [] !== $fields &&
>     $constraint &&
>     $constraint->payload &&
>     // If some or all payload fields are whitelisted, add them
>     $payloadFields = null === $fields || true === $fields ? $constraint->payload : array_intersect_key($constraint->payload, $fields)
> ) {}
> ```

This is exactly the type of code I'm concerned about, so I wouldn't even call it false positive. Looking at this, it's not immediately clear to me whether it is parsed as `($payloadFields = null === $fields) || (true === $fields ? $constraint->payload : array_intersect_key($constraint->payload, $fields))` or as `$payloadFields = ((null === $fields || true === $fields) ? $constraint->payload : array_intersect_key($constraint->payload, $fields))` or as `($payloadFields = (null === $fields || true === $fields)) ? $constraint->payload : array_intersect_key($constraint->payload, $fields)`. Yes, three possible interpretations that would be syntactically valid.

> Just thinking, if we can come up with a relevant code sample, it might also be good to add a test with an arrow function ?
> 
> Arrow functions are not 100% supported in the tokenizer layer (best effort, but the fact they can share their "scope closer" with other structures make things very difficult), so I think it may be a good idea to have some tests involving arrow functions just to be on the safe side.

Okay, will have a look.

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
TimWolla: Over the years, I've come to appreciate treating all variables (to the extent where it's reasonably possible) as effectively readonly and only assigning them once and not modifying them afterwards. I find that this generally makes the code easier to grasp and reduces bugs.

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
TimWolla: Using `\in_array()` keeps related logic within a single expression.

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
jrfnl: That's fair, but that doesn't answer my question about the use of `array_merge()`.

TimWolla: It does, because the alternative would be using `[ ...Tokens::$booleanOperators, \T_INLINE_THEN, \T_INLINE_ELSE ]` which is not legal in PHP 5.5, explicitly spelling out everything or using `$this->previousTokens[]` which would mutate `previousTokens` (and thus was excluded by the previous answer).

jrfnl:
> This code is not syntactically valid, as the || is directly followed by ) => substr($value, 1, -1), // quoted string. I did not yet test the sniff with syntactically invalid code. Did you accidentally remove some bits while copying the code into the issue?

Sorry, my bad. This is the correct code sample for the false positive (I copied a bit too much when I added the comments):

```php
        return match (true) {
            \defined($value) => \constant($value),
            'yes' === $lowercaseValue,
            'on' === $lowercaseValue => true,
            'no' === $lowercaseValue,
            'off' === $lowercaseValue,
            'none' === $lowercaseValue => false,
            isset($value[1]) && (
                ("'" === $value[0] && "'" === $value[\strlen($value) - 1]) ||
                ('"' === $value[0] && '"' === $value[\strlen($value) - 1])
            ) => substr($value, 1, -1), // quoted string
            default => XmlUtils::phpize($value),
        };
```

jrfnl:
> This is exactly the type of code I'm concerned about, so I wouldn't even call it false positive. Looking at this, it's not immediately clear to me whether it is parsed as `($payloadFields = null === $fields) || (true === $fields ? $constraint->payload : array_intersect_key($constraint->payload, $fields))` or as `$payloadFields = ((null === $fields || true === $fields) ? $constraint->payload : array_intersect_key($constraint->payload, $fields))` or as `($payloadFields = (null === $fields || true === $fields)) ? $constraint->payload : array_intersect_key($constraint->payload, $fields)`. Yes, three possible interpretations that would be syntactically valid.

Except there isn't really room for interpretation due to the `$payloadFields =` and `=` having higher precedence than any of the operators this sniff is looking for.

Still I agree that this kind of code can be confusing.

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
jrfnl: Actually, no, there is one more alternative you seem to overlook (which complies with your "don't mutate" principle):
```php
        $this->previousTokens = Tokens::$booleanOperators + [\T_INLINE_THEN => \T_INLINE_THEN, \T_INLINE_ELSE => \T_INLINE_ELSE];
```

jrfnl: Okay, so maybe I should have made the question more specific: why `array_merge()` instead of array union ?

TimWolla: The semantics of applying the `+` operator to arrays (“Array Union”) are so terrible (especially with regard to arrays that satisfy `array_is_list()`) that I don't believe there is ever a good reason to use it, because it makes the code less clear.

jrfnl: Okay, so I fundamentally disagree with you on this. Array merge and array union both have a valid use case if used correctly.

Here's my explanation of why using `array_merge()` in the context of token arrays is an anti-pattern:

> As sniffs are generally called thousands of times within the course of a PHPCS scan, performance is really important for sniffs.
> 
> When a token array has the token constants as both the keys as well as the values, it can be safely used with `isset()` when doing your own token walking in a sniff, and it can also safely be passed to the `File::find[Previous|Next]()` methods as well as various other utility methods which expect a token array with the token constants either as keys or as values.
> 
> Also keep in mind that the PHP native tokens translate to integers, while the PHPCS native tokens translate to strings. This means both can be safely used as array keys.
> 
> When combining token arrays, there are basically two (four) options:
> 1. `array_merge()`
> 2. Array union using the `+` or `+=` operator.
> 3. Adding to the array using `$array[] = TOKEN_CONSTANT` (without setting a key).
> 4. Adding to the array using `$array[TOKEN_CONSTANT] = TOKEN_CONSTANT` (setting the key to the token constant).
> 
> In my opinion, the only acceptable options in the context of combining token arrays in PHPCS are option 2 and 4.
> 
> `array_merge()` will keep string keys, but will renumber integer keys, which means that you can only rely on the _values_ in the array after the merge as the _keys_ are no longer reliable as token constants.
`array_merge()` will also overwrite the values for string keys if the value would be different in the second array (which should never be the case for token arrays).
> On top of that, for integer keys with integer values (PHP native token constants), a new entry will be added for each occurance, which leads to duplicate values in the resulting array.
> As a consequence of this, you will always need to use `in_array()`, instead of the faster `isset()`, to verify if a token exists in an array and `in_array()` is error prone due to the default of the `$strict` parameter being `false`.
> 
> In contrast, array union will keep all existing keys and values and only add non-duplicate key/value pairs. While in other contexts, this may be dangerous, in the context of token arrays which consistently have the token constant as both the key and the value, this is perfectly safe and allows for use with both `isset()` as well as `in_array()`, giving more flexibility.
> Additionally, token array combinations created like that, will never contain duplicate values.
> 
> So, while `array_merge()` is now called only once and the resulting array is only passed to `File::findPrevious()`, the current use of `array_merge()` is not necessarily problematic, it is still an anti-patterns which I don't want to see in sniffs in this code base.

jrfnl: FYI: I will add a new token collection to PHPCSUtils in the 1.1.0 version for the ternary operator tokens.

TimWolla: > Okay, so I fundamentally disagree with you on this. Array merge and array union both have a valid use case if used correctly.

I'm not surprised that exposure to different code-bases will cause folks to have different opinions with regard to how stuff should or should not look like. I'm generally dealing with either lists (i.e. numbered indices from 0 to n without holes) or maps (i.e. string keys) and in those cases `array_merge()` does exactly the right thing.

> it is still an anti-patterns which I don't want to see in sniffs in this code base.

Yeah, I'm most certainly not here to discuss your coding style (if you wouldn't have explicitly asked your question) and have now shamelessly stolen the desired style from #277.

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
TimWolla: Please resolve this if no change is requested / if my response is sufficiently satisfying.

TimWolla:
You already agreed, so all is fine, but I wanted to note this nonetheless:

> Except there isn't really room for interpretation due to the `$payloadFields =` and `=` having higher precedence than any of the operators this sniff is looking for.

I actually needed to print the AST using https://github.com/nikic/PHP-Parser to be sure how it was interpreted.

In fact the fact that

```
$a && $d = null === $b || true === $c ? true : false;
```

is 

```
$a && ($d = ((null === $b || true === $c) ? true : false));
```

but 

```
$a && null === $b || true === $c ? true : false;
```

is

```
(($a && null === $b) || true === $c) ? true : false;
```

is surprising to me.

jrfnl:
> You already agreed, so all is fine, but I wanted to note this nonetheless:
> 
> > Except there isn't really room for interpretation due to the `$payloadFields =` and `=` having higher precedence than any of the operators this sniff is looking for.
> 
> I actually needed to print the AST using https://github.com/nikic/PHP-Parser to be sure how it was interpreted.
> 
> In fact the fact that
> 
> ```
> $a && $d = null === $b || true === $c ? true : false;
> ```
> 
> is
> 
> ```
> $a && ($d = ((null === $b || true === $c) ? true : false));
> ```
> 
> but
> 
> ```
> $a && null === $b || true === $c ? true : false;
> ```
> 
> is
> 
> ```
> (($a && null === $b) || true === $c) ? true : false;
> ```
> 
> is surprising to me.

Guess, you never had the chance to see my "The big "Why equal doesn't equal" Quiz" ? 😂 

But yes, operator precedence can be a bitch and I see way too many issues created because people don't understand/know it well enough (which is all the more reason why I'm happy to accept this sniff in this package).

You already saw PR #277 and yes, that was fun to write what with the operator precedence issues which could be caused when auto-fixing ;-)

TimWolla:
> Sorry, my bad. This is the correct code sample for the false positive (I copied a bit too much when I added the comments):

Simplified:

```
<?php

match (true) {
	$a || ($b && $c) => true,
};
```

This is somehow caused by the `match()` behaving differently than e.g. an array:

```
<?php

[
	$a || ($b && $c) => true,
];
```

Specifically `findStartOfStatement` returns the `$a` for both operators in the `match()` and `$a` for the `||` and `$b` for the `&&` in the array.

While I don't want to blame the upstream library, I must say that this is very surprising behavior for constructs that are so similar. Is this a bug in CodeSniffer itself or is this intentional? If this is intentional, do you have some specific advice / code snippet on how to fix this in this Sniff, ideally without replicating the entire logic?

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
```diff
@@ -0,0 +1,98 @@
...
+            // Identical operator found.
+            || $tokens[$previous]['code'] === $tokens[$stackPtr]['code']
+            // Beginning of the expression found for the ternary conditional operator.
+            || \in_array($tokens[$previous]['code'], [\T_INLINE_THEN, \T_INLINE_ELSE], true)
+        ) {
+            return;
+        }
+
+        // We found a mismatching operator, thus we must report the error.
+        $error = 'Mixing different binary boolean operators within an expression without using parentheses to clarify precedence is not allowed.';
```
TimWolla: How to break this line to not exceed the maximum line length?

Universal/Docs/CodeAnalysis/MixedBooleanOperatorStandard.xml
```diff
@@ -0,0 +1,39 @@
...
+$one = false;
+$two = false;
+$three = true;
+
+$result = <em>($one && $two) || $three</em>;
+$result2 = <em>$one && ($two || $three)</em>;
+$result3 = <em>($one && !$two) xor $three;</em>;
+$result4 = <em>$one && (!$two xor $three);</em>;
+
+if (<em>($result && !$result3) || (!$result && $result3)</em>) {}
```
jrfnl: This line will break out of the width allowed for the code samples. Same goes for the "invalid" `if` code sample.

These will need to be split across multiple lines to stay within the allowed width.

image

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
jrfnl: ☝🏻

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
jrfnl: ☝🏻

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
jrfnl: I'm fine leaving this as-is for now. I'll switch it to an `isset()` once the `Collections::ternaryOperators()` method is available in PHPCSUtils.

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
jrfnl: Yeah, that's one of the awkward things from PSR-12:
> Lines SHOULD NOT be longer than 80 characters; lines longer than that SHOULD be split into multiple subsequent lines of no more than 80 characters each.

While not pretty, I suggest doing what PSR-12 proscribes: splitting the line.

jrfnl (commented):
I've given the sniff another look over and left some comments, but it looks like a number of my remarks from my [previous comment](https://github.com/PHPCSStandards/PHPCSExtra/pull/271#pullrequestreview-1634348864) haven't been addressed yet:
* Deciding on a final name for the sniff.
* Deciding whether to include the "binary" terminology in the messaging. I'm unsure whether that would clarify things or cause more confusion.
* "Boolean operators" vs "logical operators" terminology.
* Adding some tests which involve arrow functions.


> While I don't want to blame the upstream library, I must say that this is very surprising behavior for constructs that are so similar. Is this a bug in CodeSniffer itself or is this intentional? If this is intentional, do you have some specific advice / code snippet on how to fix this in this Sniff, ideally without replicating the entire logic?

Looks like this was done [by design in PHPCS upstream](https://github.com/squizlabs/PHP_CodeSniffer/commit/ef80e53de0e3c78f1a5599310d2b9803061ae9bd), though I'm not sure that was the [correct choice](https://github.com/squizlabs/PHP_CodeSniffer/blob/master/src/Files/File.php#L2375-L2423). Having said that, I don't think comparing match conditions with array keys is an even remotely fair comparion. Array keys are always singular (one key per item), while match branches can have multiple comma-separated conditions, which, what with the comma also being the end of the "return" statement makes parsing `match` expressions based on the tokens a lot harder.

If needs be, I'd be okay with adding the simplified `match` code to the tests and marking this as a known false positive + opening an issue to follow up on this later (and adding a link to the issue as a comment to the test).

TimWolla:
> I've given the sniff another look over and left some comments, but it looks like a number of my remarks from my [previous comment](https://github.com/PHPCSStandards/PHPCSExtra/pull/271#pullrequestreview-1634348864)

Yes, I plan to do the renaming and related changes once the functionality is clear and correct to avoid unnecessary churn. I was waiting for advice regarding the match statement, before proceeding. All the stuff that is not yet collapsed still is on my list.

Universal/Sniffs/CodeAnalysis/MixedBooleanOperatorSniff.php
TimWolla: I was specifically asking *how* you prefer to split this. Is something like this:

```
$error = 'Mixing different binary boolean operators within an '
    . 'expression without using parentheses to clarify precedence is not allowed.';
```

what you expect to see?

jrfnl: I'm not too fussed about that, either what you propose or something along the lines of the below should be fine.

```php
$error  = 'Mixing different binary boolean operators within an expression';
$error .= ' without using parentheses to clarify precedence is not allowed.';
```

There's bound to be "prior art" in this repo for these kind of splits, but I'm honestly not sure if I have been consistent with these.

Universal/Docs/CodeAnalysis/MixedBooleanOperatorStandard.xml
TimWolla: Thank you for that command, that was helpful. I've also inserted additional line breaks to make sure both versions line up properly.

TimWolla:
Okay, I managed to fix the handling for the match case and added arrow function tests.

I believe this should now be everything with regard to the implementation. I'll think about the naming adjustments once you confirm the implementation is correct and complete.

Do I still need “syntax error” tests or did that become irrelevant with the simplified implementation?

jrfnl:
> Okay, I managed to fix the handling for the match case and added arrow function tests.

@TimWolla Thank you for making those changes and updates. I've tried to come up with cases which could break the current implementation, but my imagination struck out, so I think we're good. Well done! 🎉 

Might be worth adding the following extra tests still to safeguard things further, but these are already handled correctly:
```php
// Not ok.
match (true) {
    $a || ($b && $c) && $d => true,
    $b && $c['a'] || $d => true,
    $b && ${$var} || $d => true,
};
```

> I believe this should now be everything with regard to the implementation. I'll think about the naming adjustments once you confirm the implementation is correct and complete.

I'm happy with the implementation as it is now. Let's get the name sorted & get this merged!

> Do I still need “syntax error” tests or did that become irrelevant with the simplified implementation?

With the current implementation, I don't see a case where a parse error could cause the sniff to throw a PHP notice, so as far as I'm concerned it's not needed for the sniff as-it-is now.

TimWolla:
Okay, I:

- Added the additional test case.
- Explained why `!` is not handled.
- Adjusted the copyright line.
- Added the requested references to other sniffs.
- Renamed the Sniff to `RequireExplicitBooleanOperatorPrecedence`. I've opted for “Boolean” instead of “Logical”, because:
   - The internal token array is called `$booleanOperators`.
   - It's mainly about `&&` and `||` (because no one should use `and`, `or`, `xor` in the first place) and the internal tokens are `T_BOOLEAN_AND`, `T_BOOLEAN_OR` vs `T_LOGICAL_AND`, `T_LOGICAL_OR`, `T_LOGICAL_XOR` for the others.
   - PSR-12 uses “boolean operator”.
   - It matches the name of PSR12.ControlStructures.BooleanOperatorPlacementSniff, which is reasonably related, because operator placement and precedence go hand in hand.

TimWolla:
Will squash the entire commit history once you approve that everything is good.

jrfnl:
FYI: I've moved this PR out of the 1.2.0 milestone as, what with the [PHPCS repo move](https://github.com/squizlabs/PHP_CodeSniffer/issues/3932), the sniff can go into the main [PHPCS repo](https://github.com/PHPCSStandards/PHP_CodeSniffer) as soon as its been pulled there (as it's been extensively reviewed here already), which is what the Op indicated originally as the preferred option. I've discussed this with @TimWolla off-GH.

PR #276, opened 2023-09-21
:sparkles: New `Universal.PHP.LowercasePHPTag` sniff

jrfnl:
... to enforce that the "PHP" in a PHP open tag is lowercase.

Includes fixer.
Includes unit tests.
Includes documentation.
Includes metrics.

### DIFF
--- a/Universal/Sniffs/PHP/LowercasePHPTagSniff.php
+++ b/Universal/Sniffs/PHP/LowercasePHPTagSniff.php
@@ -1,0 +1,87 @@
+<?php
+/**
+ * PHPCSExtra, a collection of sniffs and standards for use with PHP_CodeSniffer.
+ *
+ * @package   PHPCSExtra
+ * @copyright 2020 PHPCSExtra Contributors
+ * @license   https://opensource.org/licenses/LGPL-3.0 LGPL3
+ * @link      https://github.com/PHPCSStandards/PHPCSExtra
+ */
+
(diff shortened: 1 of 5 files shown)


Universal/Sniffs/PHP/LowercasePHPTagSniff.php
```diff
@@ -0,0 +1,87 @@
...
+        } else {
+            $errorCode = 'Mixedcase';
+            $phpcsFile->recordMetric($stackPtr, self::METRIC_NAME, 'mixed case');
+        }
+
+        $fix = $phpcsFile->addFixableError(
+            'The php open tag should be in lowercase. Found: %s',
+            $stackPtr,
+            $errorCode,
+            [$content]
```
fredden: Should we remove white-space from `$content` in the message here? Running the sniff over the test file gives an "empty" row in the report table after line 4. The trailing spaces for lines 7 & 8 are less obvious.

Screenshot_2023-10-05_00-14-07

```suggestion
            [\trim($content)]
```

fredden (approved):
This sniff looks good. I've tested this locally. As far as I can tell, this behaves as expected.
:+1:

Universal/Sniffs/PHP/LowercasePHPTagSniff.php
jrfnl: Good point! Thanks. Fixed now.

fredden (approved):
Re-reviewed and tested locally.

PR #277, opened 2023-09-21
:sparkles: New `Universal.CodeAnalysis.NoDoubleNegative` sniff

jrfnl:
... to detect double negatives (`!!`) and advise to use a boolean cast instead.

The sniff will correctly handle a situation where even more consecutive not operators are found.

In the case, the number of operators is uneven, it will auto-fix to a single not operator.

And when a double not operator is found before an expression involving `instanceof`, the error will still be thrown, but not auto-fix as the `instanceof` operator is nested right between the `!` operator and a `(bool)` cast operator precedence wise, so auto-fixing without also adding parentheses would change the behaviour of the code.

Includes fixer.
Includes unit tests.
Includes documentation.

Ref: https://www.php.net/manual/en/language.operators.precedence.php

### DIFF
--- a/Universal/Sniffs/CodeAnalysis/NoDoubleNegativeSniff.php
+++ b/Universal/Sniffs/CodeAnalysis/NoDoubleNegativeSniff.php
@@ -1,0 +1,269 @@
+<?php
+/**
+ * PHPCSExtra, a collection of sniffs and standards for use with PHP_CodeSniffer.
+ *
+ * @package   PHPCSExtra
+ * @copyright 2023 PHPCSExtra Contributors
+ * @license   https://opensource.org/licenses/LGPL-3.0 LGPL3
+ * @link      https://github.com/PHPCSStandards/PHPCSExtra
+ */
+
+namespace PHPCSExtra\Universal\Sniffs\CodeAnalysis;
+
+use PHP_CodeSniffer\Files\File;
+use PHP_CodeSniffer\Sniffs\Sniff;
+use PHP_CodeSniffer\Util\Tokens;
+use PHPCSUtils\BackCompat\BCFile;
+use PHPCSUtils\Utils\GetTokensAsString;
+use PHPCSUtils\Utils\Parentheses;
+
+/**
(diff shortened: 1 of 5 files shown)


Universal/Sniffs/CodeAnalysis/NoDoubleNegativeSniff.php
```diff
@@ -0,0 +1,269 @@
...
+     */
+    public function register()
+    {
+        // Collect all the operators only once.
+        $this->operatorsWithLowerPrecedence                 = Tokens::$assignmentTokens;
+        $this->operatorsWithLowerPrecedence                += Tokens::$booleanOperators;
+        $this->operatorsWithLowerPrecedence                += Tokens::$comparisonTokens;
+        $this->operatorsWithLowerPrecedence                += Tokens::$operators;
+        $this->operatorsWithLowerPrecedence[\T_INLINE_THEN] = \T_INLINE_THEN;
+        $this->operatorsWithLowerPrecedence[\T_INLINE_ELSE] = \T_INLINE_ELSE;
```
diedexx, line 58: Out of curiosity: why not use the constructor for this?

diedexx (approved):
I wasn't able to test to confirm it, but this might be a case that isn't detecting the double `!`. 
```
echo !@! $undefined_var;
```
Apparently you can silence the undefined variable warning in this way. Not sure why anyone would though. 🤷

Universal/Sniffs/CodeAnalysis/NoDoubleNegativeSniff.php
jrfnl: It's rare for a sniff to have a constructor and the `register()` method is normally only called once anyway, so as good a place as any to do this in.

jrfnl:
> I wasn't able to test to confirm it, but this might be a case that isn't detecting the double `!`.
> 
> ```
> echo !@! $undefined_var;
> ```
> 
> Apparently you can silence the undefined variable warning in this way. Not sure why anyone would though. 🤷

Good point and no, the sniff would not flag that, but I'm not sure the sniff _should_ catch that...

PR #284, opened 2023-11-07
Split the errors, adding the *CloserSameLine ones

### LINKED ISSUE #283: Problem with NormalizedArrays.Arrays.CommaAfterLast in some cases
## Bug Description
The `NormalizedArrays.Arrays.CommaAfterLast` does its job pretty well. But there is a case that is hitting us a lot (because of our own coding standard explicitly allowing it), and that I think that can be hitting others out there too.

Right now we support all these arrays as valid:
```
$a = [1, 2, 3, 4]; // Normal mono-line.
$a = [
    1,
    2,
    3,
    4,
]; // Normal multi-line
$a = [1, 2,
    3, 4]; // Multi-multi line with same line closer (to call it some way).
$a = [
    1, 2,
    3, 4,
]; // Multi-multi line with closer apart (also to call it some way).
```
And the Sniff is working ok for all the cases but the 3rd one. And it seems logic not to require a comma there, because, if there are new items to be added to the array later, the line is going to change, yes or yes (the closer will need to be moved.

## Given the following reproduction Scenario
Ensure that "Multi-multi line with same line closer " cases don't require a comma.

The issue happens when running this command:

```bash
vendor/bin/phpcs --standard=NormalizedArrays --sniffs=NormalizedArrays.Arrays.CommaAfterLast test.php
```

... over a file containing this code:
```php
<?php

$a = [1, 2,
    3, 4];

$a = [1 => 1, 2 => 2,
    3 => 3, 4 => 4];

$a = [1 => [1 => 1,
    2 => 2,
    3 => 3,
    4 => 4,
], 5 => 5];

$a = [1 => [1 => 1,
    2 => 2,
    3 => 3,
    4 => 4],
5 => 5];
```

### I'd expect the following behaviour
No errors are reported.

### Instead this happened
Five errors are reported.
```
-------------------------------------------------------------------------------------------
FOUND 5 ERRORS AFFECTING 5 LINES
-------------------------------------------------------------------------------------------
  4 | ERROR | [x] There should be a comma after the last array item in a multi-line array.
  7 | ERROR | [x] There should be a comma after the last array item in a multi-line array.
 13 | ERROR | [x] There should be a comma after the last array item in a multi-line array.
 18 | ERROR | [x] There should be a comma after the last array item in a multi-line array.
 19 | ERROR | [x] There should be a comma after the last array item in a multi-line array.
-------------------------------------------------------------------------------------------
PHPCBF CAN FIX THE 5 MARKED SNIFF VIOLATIONS AUTOMATICALLY
-------------------------------------------------------------------------------------------
```

## Environment

| Environment             | Answer                                                                         |
|-------------------------|--------------------------------------------------------------------------------|
| PHP version             | Any                                                                         |
| PHP_CodeSniffer version | 3.7.2                                                                          |
| PHPCSExtra version      | 1.1.1                                                                         |
| PHPCSUtils version      | 1.0.8                                                                         |
| Install type            | Normal composer, used by [our standard ](https://github.com/moodlehq/moodle-cs/blob/main/composer.json)|

## Additional Context (optional)
This was originally reported for us @ https://github.com/moodlehq/moodle-cs/issues/76
And this change seems to fix the problem:
```
diff --git a/NormalizedArrays/Sniffs/Arrays/CommaAfterLastSniff.php b/NormalizedArrays/Sniffs/Arrays/CommaAfterLastSniff.php
index d03d1ff..6472e5c 100644
--- a/NormalizedArrays/Sniffs/Arrays/CommaAfterLastSniff.php
+++ b/NormalizedArrays/Sniffs/Arrays/CommaAfterLastSniff.php
@@ -159,6 +159,12 @@ final class CommaAfterLastSniff implements Sniff
                     return;
                 }
 
+                // If the line of the last non-empty token is the same as the closer, we're not
+                // going to require a comma. TODO: Put this under some Sniff configuration setting.
+                if ($tokens[$lastNonEmpty]['line'] === $tokens[$closer]['line']) {
+                    return;
+                }
+
                 $error     = 'There should be a comma after the last array item in a %s array.';
                 $errorCode = 'Missing' . $errorCode;
                 $data      = [$phrase];
```
Note the TODO that I've put there. IMO there are 4 alternatives:
- We make that the default behaviour. After all, it doesn't make sense to force commas when the closer is in the same line.
- We make it a new `$validValues`, say: `enforce_but_when_sameline_closer`.
- We make it a new, apart sniff setting: `enforce_skipped_when_sameline_closer`.
- The change is not acceptable (so we'll have to duplicate the sniff into our standard and apply the patch there.

If there is any insight/preference about any of the 1-3 alternatives above... I'm happy preparing a PR with tests and so on...

Ciao :-)

## Tested Against `develop` branch?
- [x] I have verified the issue still exists in the `develop` branch of PHPCSExtra.

stronk7:
Some standards may want to have different rules when the array closer is in the same line than the last element.

When that happens, the new *CloserSameLine errors are emitted, allowing to disable them via ruleset.

For testing, a couple of cases have been added to CommaAfterLastUnitTest.1.inc and then, the fixtures have been 100% duplicated into CommaAfterLastUnitTest.3.inc that runs with the *CloserSameLine errors disabled.

A simple diff shows the 8 cases in which the outcome is different, as expected.

Fixes #283.

### DIFF
--- a/NormalizedArrays/Sniffs/Arrays/CommaAfterLastSniff.php
+++ b/NormalizedArrays/Sniffs/Arrays/CommaAfterLastSniff.php
@@ -145,6 +145,13 @@ final class CommaAfterLastSniff implements Sniff
             return;
         }
 
+        // If the closer is on the same line as the last element, change the error code for multi-line arrays.
+        if ($errorCode === 'MultiLine'
+            && $tokens[$lastNonEmpty]['line'] === $tokens[$closer]['line']
+        ) {
+            $errorCode .= 'CloserSameLine';
+        }
+
         $isComma = ($tokens[$lastNonEmpty]['code'] === \T_COMMA);
 
         $phpcsFile->recordMetric(
--- a/NormalizedArrays/Tests/Arrays/CommaAfterLastUnitTest.1.inc
+++ b/NormalizedArrays/Tests/Arrays/CommaAfterLastUnitTest.1.inc
@@ -166,6 +166,56 @@ EOD
 , /*comment*/
 ) );
 
+/**
+ * Tests enforcing a comma after the last array item when the closer is in the same line. See #283.
+ */
+// phpcs:set NormalizedArrays.Arrays.CommaAfterLast multiLine enforce
+
+$missing = array(
+    1, 2,
+    3, 4);
+
+$missing = [
+    '1', '2',
+    '3', '4'];
+
+$missing_but_good = [
+    '1', '2',
+    '3', '4']; // phpcs:ignore NormalizedArrays.Arrays.CommaAfterLast.MissingMultiLineCloserSameLine
+
+$good = array(
+    1, 2,
+    3, 4,);
+
+$good = [
+    '1', '2',
+    '3', '4',];
+
+/**
+ * Tests forbidding a comma after the last array item when the closer is in the same line. See #283.
+ */
+// phpcs:set NormalizedArrays.Arrays.CommaAfterLast multiLine forbid
+
+$found = array(
+    1, 2,
+    3, 4,);
+
+$found = [
+    '1', '2',
+    '3', '4',];
+
+$found_but_good = [
+    '1', '2',
+    '3', '4',]; // phpcs:ignore NormalizedArrays.Arrays.CommaAfterLast.FoundMultiLineCloserSameLine
+
+$good = array(
+    1, 2,
+    3, 4);
+
+$good = [
+    '1', '2',
+    '3', '4'];
+
 // Reset the properties to the defaults.
 // phpcs:set NormalizedArrays.Arrays.CommaAfterLast singleLine forbid
 // phpcs:set NormalizedArrays.Arrays.CommaAfterLast multiLine enforce
--- a/NormalizedArrays/Tests/Arrays/CommaAfterLastUnitTest.1.inc.fixed
+++ b/NormalizedArrays/Tests/Arrays/CommaAfterLastUnitTest.1.inc.fixed
@@ -166,6 +166,56 @@ EOD
  /*comment*/
 ) );
 
+/**
+ * Tests enforcing a comma after the last array item when the closer is in the same line. See #283.
+ */
+// phpcs:set NormalizedArrays.Arrays.CommaAfterLast multiLine enforce
+
+$missing = array(
+    1, 2,
+    3, 4,);
+
+$missing = [
+    '1', '2',
+    '3', '4',];
+
+$missing_but_good = [
+    '1', '2',
+    '3', '4']; // phpcs:ignore NormalizedArrays.Arrays.CommaAfterLast.MissingMultiLineCloserSameLine
+
+$good = array(
+    1, 2,
+    3, 4,);
+
+$good = [
+    '1', '2',
+    '3', '4',];
+
+/**
+ * Tests forbidding a comma after the last array item when the closer is in the same line. See #283.
+ */
+// phpcs:set NormalizedArrays.Arrays.CommaAfterLast multiLine forbid
+
+$found = array(
+    1, 2,
+    3, 4);
+
+$found = [
+    '1', '2',
+    '3', '4'];
+
+$found_but_good = [
+    '1', '2',
+    '3', '4',]; // phpcs:ignore NormalizedArrays.Arrays.CommaAfterLast.FoundMultiLineCloserSameLine
+
+$good = array(
+    1, 2,
+    3, 4);
+
+$good = [
+    '1', '2',
+    '3', '4'];
+
 // Reset the properties to the defaults.
 // phpcs:set NormalizedArrays.Arrays.CommaAfterLast singleLine forbid
 // phpcs:set NormalizedArrays.Arrays.CommaAfterLast multiLine enforce
--- a/NormalizedArrays/Tests/Arrays/CommaAfterLastUnitTest.php
+++ b/NormalizedArrays/Tests/Arrays/CommaAfterLastUnitTest.php
@@ -52,6 +52,10 @@ final class CommaAfterLastUnitTest extends AbstractSniffUnitTest
                     152 => 1,
                     159 => 1,
                     166 => 1,
+                    176 => 1,
+                    180 => 1,
+                    201 => 1,
+                    205 => 1,
                 ];
 
             case 'CommaAfterLastUnitTest.2.inc':


stronk7:
OT, sharing because it happened to me... I just found that something in composer is not installing things properly, neither `vendor/bin/phpunit` or `composer test` (and others) seem to be working by default. Had to remove the lock file, remove the platform constraint and other custom installations to get it working locally.

jrfnl:
@stronk7 Thanks for the PR, I'll try to review it over the next few days.

> OT, sharing because it happened to me... I just found that something in composer is not installing things properly, neither `vendor/bin/phpunit` or `composer test` (and others) seem to be working by default. Had to remove the lock file, remove the platform constraint and other custom installations to get it working locally.

Just tried myself and all seems fine, so I'm not sure what happened for you.
Removing the `lock` file and (re-)running `composer update` (on PHP 7.4 or lower) is never a bad idea though if you cloned a while back.
There is no committed `composer.lock` file, so that is normal.

As for the "platform constraint": the repo doesn't have one, so I don't know what you mean.

The only real (known) problem with the tests is that you need to run them locally on PHP 7.4 or lower as otherwise you run into issues with PHPUnit 7.5. There are work-arounds for that, see: https://github.com/PHPCSStandards/PHPCSExtra/blob/develop/.github/workflows/quicktest.yml , but the simple solution is to just run on PHP 7.4.
This is a limitation in the PHPCS native test suite which is used for the tests and I'm [trying to get that limitation lifted](https://github.com/squizlabs/PHP_CodeSniffer/pull/3803).

jrfnl:
@stronk7 I've just had a quick glance at the PR and first thing I noticed is that the new error code is not just reserved for multi-line arrays.
This will need to be fixed, as differentiating the error code doesn't make sense for a single line array where the closer will always be on the same line and it would basically mean the `FoundSingleLine` and `MissingSingleLine` codes would be replaced with the new codes which is a much bigger and unnecessary breaking change.

jrfnl:
P.S.: no need to worry about the code coverage builds failing - this is [a known issue and I'm working with Coveralls to fix that](https://github.com/lemurheavy/coveralls-public/issues/1721).

stronk7:
Thanks for the composer comments and tricks… I’ll try again tomorrow from scratch. Maybe I had some old lock file around and that caused the troubles…

Regarding applying it only to multilines, cannot but agree. As commented in the issue, it really doesn’t make sense for singles. I will amend the PR tomorrow, thanks!

stronk7:
Patch updated to only apply the "CloserSameLine" error code suffix to multi lines (with tests amended to ensure that's what it's being done).

Ciao :-)

NormalizedArrays/Sniffs/Arrays/CommaAfterLastSniff.php
```diff
@@ -147,6 +147,12 @@ public function process(File $phpcsFile, $stackPtr)
 
         $isComma = ($tokens[$lastNonEmpty]['code'] === \T_COMMA);
 
+        // If the closer in the same line than the last element, change the error code for multi-line arrays.
```
jrfnl:
```suggestion
        // If the closer is on the same line as the last element, change the error code for multi-line arrays.
```

NormalizedArrays/Sniffs/Arrays/CommaAfterLastSniff.php
```diff
@@ -147,6 +147,12 @@ public function process(File $phpcsFile, $stackPtr)
...
+        $sameLineCloser = ($tokens[$lastNonEmpty]['line'] === $tokens[$closer]['line']);
+        if ($errorCode === 'MultiLine' && $sameLineCloser) {
+            $errorCode .= 'CloserSameLine';
+        }
```
jrfnl:
```suggestion
        if ($errorCode === 'MultiLine'
            && $tokens[$lastNonEmpty]['line'] === $tokens[$closer]['line']
        ) {
            $errorCode .= 'CloserSameLine';
        }
```

(no need to declare a variable if the value is only used once)

I'd also suggest moving this block up to before the `$isComma` (and closer to the other error code logic).

jrfnl (changes requested):
@stronk7 Hi Eloy, thank you for fixing that up.

I've now done a proper review of the PR and the principle of the fix is now correct 👍🏻 .
I've left some nitpicky comments inline to tidy that up a little still, but that doesn't affect the actual functional change, which is good.

As for the tests...
While I appreciate how much attention you've given to the tests, the way this is done now, means the diff is much bigger than it should be, what with all test lines being renumbered. This also makes it much harder than it should be to review the PR.

As a rule of thumb: always put new tests at the _bottom_ of a test case file (or in a new file if so warranted) with a comment explaining what those tests are related to.
Do not touch or cause to re-number the pre-existing tests.

In all honesty, the `3` test case file is redundant, as it isn't testing anything new in the PHPCSExtra repo. It is testing whether PHPCS handles the disable/ignore annotations correctly, but that is something which should be (and is) tested in the PHPCS repo.

So, in short:
* Please remove the `3` test case file.
* Please remove the comment at the top of the `1` test case file.
* Please move the new tests in the `1` test case file to the bottom of the file (just above the reset) and add appropriate `phpcs:set` comments for each situation and add a comment to describe why these tests were added and what is specifically being tested with them. Include the issue number in the comment as a reference.

As for the actual tests, these are only testing the "unhappy" path (= the path which throws an error).
I'm missing tests with multi-line arrays with the closer on the same line as the last content, which _don't_ trigger an error, both for the `forbid`, as well as the `enforce` case.

And for completeness of testing the actual change, you _may_ want to also add single line test code just to demonstrate that those aren't affected, but I'm fine with leaving those out as well, as this can also be confirmed via the pre-existing tests.

Sorry to be a pain. Please let me know if you have questions.

jrfnl:
P.S.: if you rebase the PR on the current `develop`, the code coverage builds should start working ;-)

stronk7:
> Sorry to be a pain. Please let me know if you have questions.

No worries, I'm used, LOL :-)

Thanks for the review, I'll go towards all the the test changes and suggestions later today.

stronk7:
Ok, I think I've followed all your suggestions and the new candidate PR follows all them.

Only point that I'm not 100% convinced is the one about not covering with tests what happens when the `CloserSameLine` error is disabled. While I agree that it's `phpcs` itself the one to make the disabling to work... it's the only way we have to ensure that we are setting the `CloserSameLine` new error properly. I mean, current tests don't cover the new code introduced in the Sniff. At all (in fact they will pass exactly the same without the Sniff changes).

Note I don't mind much, because for sure we are going to add some tests covering that in our standard, but I think that it should be tested here because, as said, it's not covering the new real code at all.

Ciao :-)

stronk7:
Adding on my previous comment... what I'd add to the tests is this (I think it's the minimum patch I've been able to create):

https://github.com/stronk7/PHPCSExtra/commit/f971d5b0e74f44daff633119ebd2b87fa85f4618

You can see that it doesn't change the tests outcome (all the new lines are ok, because we are ignoring the `CloserSameLine`error) but, that way, we are getting covered the new Sniff code.

For your consideration, I'm happy to fixup this PR with that commit or to leave it apart (and to add a similar testing to our standard, ensuring that the new error codes work as expected).

Ciao :-)

jrfnl:
@stronk7 

> Only point that I'm not 100% convinced is the one about not covering with tests what happens when the `CloserSameLine` error is disabled. While I agree that it's `phpcs` itself the one to make the disabling to work... it's the only way we have to ensure that we are setting the `CloserSameLine` new error properly. I mean, current tests don't cover the new code introduced in the Sniff. At all (in fact they will pass exactly the same without the Sniff changes).
> 
> Note I don't mind much, because for sure we are going to add some tests covering that in our standard, but I think that it should be tested here because, as said, it's not covering the new real code at all.

> Adding on my previous comment... what I'd add to the tests is this (I think it's the minimum patch I've been able to create):
> 
> https://github.com/stronk7/PHPCSExtra/commit/f971d5b0e74f44daff633119ebd2b87fa85f4618
> 
> You can see that it doesn't change the tests outcome (all the new lines are ok, because we are ignoring the `CloserSameLine`error) but, that way, we are getting covered the new Sniff code.

Okay, so let's split this up:
1. The actual code does get hit via these tests (also see the code coverage), so the code does actually get tested.
2. The tests wouldn't fail without this change.

I agree that 2 is problematic, and that's inherent to the PHPCS native test framework, which is being used as a base for these tests.
It's on my never-ending to-do list to create an alternative test setup for external standards, which would allow for testing that the error codes are as expected (and more). Also see: https://github.com/PHPCSStandards/PHPCSUtils/issues/72

I'd be okay with you adding the extra tests you propose. I think only one for each type (missing/found) would be fine and that those can be in the same `// phpcs:set...` blocks as the tests you've got in place now.

jrfnl (approved):

stronk7:
> I'd be okay with you adding the extra tests you propose. I think only one for each type (missing/found) would be fine and that those can be in the same `// phpcs:set...` blocks as the tests you've got in place now.

Ok, will amend the commit to incorporate ONE missing and ONE found case to the 2 "new sections" fixtures introduced here.

OT: Since we started to use phpcs long ago... one of the first things we did is to [switch to a "richer" testing alternative](https://github.com/moodlehq/moodle-cs/blob/main/moodle/Tests/MoodleCSBaseTestCase.php). It supports the normal `line => num_errors` thing of phpcs [and more](https://github.com/moodlehq/moodle-cs/blob/main/moodle/Tests/PHPIncludingFileTest.php), with both the error message and the error code being available for substring assertions. Just sharing in case it gives some idea to you (in fact I think we shared about it @ phpcs ages ago).

jrfnl:
> > I'd be okay with you adding the extra tests you propose. I think only one for each type (missing/found) would be fine and that those can be in the same `// phpcs:set...` blocks as the tests you've got in place now.
> 
> Ok, will amend the commit to incorporate ONE missing and ONE found case to the 2 "new sections" fixtures introduced here.

👍🏻 

> OT: Since we started to use phpcs long ago... one of the first things we did is to [switch to a "richer" testing alternative](https://github.com/moodlehq/moodle-cs/blob/main/moodle/Tests/MoodleCSBaseTestCase.php). It supports the normal `line => num_errors` thing of phpcs [and more](https://github.com/moodlehq/moodle-cs/blob/main/moodle/Tests/PHPIncludingFileTest.php), with both the error message and the error code being available for substring assertions. Just sharing in case it gives some idea to you (in fact I think we shared about it @ phpcs ages ago).

Looks interesting, but not quite what I had in mind. We use a [custom test case in PHPCompatibility](https://github.com/PHPCompatibility/PHPCompatibility/blob/develop/PHPCompatibility/Tests/BaseSniffTestCase.php) as well.

Basically, what I'm looking to create would be a "best of these worlds" situation, where it would give sniff writers the flexibility to test what they want to test, while also making the test suites straight-forward to maintain. Keep en eye on PHPCSUtils or subscribe to the issue if you'd like to stay informed.

stronk7:
Patch amended, by adding the 2 agreed new cases (using phpcs:ignore).

> Keep en eye on PHPCSUtils or subscribe to the issue if you'd like to stay informed.
👍

jrfnl (approved):
All good! Thanks for contributing and for bearing with me to get this to where it is now.

PR #327, opened 2025-03-01
:sparkles: New Universal.PHP.NoFQNTrueFalseNull sniff

jrfnl:
... to forbid using `true`, `false` and `null` as fully qualified constants.

Includes fixer.
Includes unit tests.
Includes documentation.

_Note: yes, there is a related tokenizer issue in PHPCS where these tokens are not tokenized the same PHP cross-version, however, the tokenization will change in PHPCS 4.0 as well, so there isn't much point addressing that now as it would just make things even more complicated for sniffs come PHPCS 4.0._

_And yes, this does mean that this sniff will need adjusting for PHPCS 4.0._


jrfnl:
@rodrigoprimo Fancy reviewing this PR ?

jrfnl:
Rebased without changes to allow for new CI check.

rodrigoprimo (approved):
Looks good to me!

(sorry for the delay to review this PR)

jrfnl:
Thanks for reviewing @rodrigoprimo !

PR #352, opened 2025-06-08
Universal/ModifierKeywordOrder: add tests with PHP 8.3 typed constants

jrfnl:
Just to safeguard the sniff does not get confused over this, but shouldn't be a problem as the sniff only looks at the tokens before the `const` keyword.


jrfnl:
Duplicate of #351 - sorry, I cherrypicked the wrong commit.

PR #360, opened 2025-06-13
GH Actions: tweak the way the CS dependency versions are set

jrfnl:
PR #214 previously already changed the logic of the workflow to take advantage of the Composer `--prefer-lowest` setting. This commit takes this yet another step further.

### Stable vs Dev

As things were, the sniff tests were being run against the _lowest_ supported CS dependencies (PHPCS + PHPCSUtils) and the _development_ version of PHPCS (but not PHPCSUtils).

While testing against the _dev_ versions of CS dependencies is relevant, it is more important that the tests safeguard that the sniffs work correctly on typical user-installable combinations of the CS dependencies.

Now, it could be argued that every single combination of the different versions of each of these dependencies should be tested, but that would make the matrix _huge_ for little added benefit.

So, I'm changing the `quicktest` and the `coverage` matrices to test by default against the `lowest` and `stable` supported versions of all dependencies, i.e low PHPCS + PHPCSUtils and high PHPCS + PHPCSUtils. This allows to simplify those workflows a little too.

Note: `stable` is also the name used for the Composer sister-argument to `--prefer-lowest`: `--prefer-stable`, so should be clear-cut for anyone reading the workflow.

## Dev and version specific testing

Now we do want to keep testing against dev versions as well and, what with PHPCS 4.0 around the corner, may want to get more specific about the versions being tested sometime soon.

By decoupling the setting of the PHPCS version versus the setting of the PHPCSUtils version, we gain that flexibility.

So, for the `test` job, I'm adding a new key `phpcsutils_version`, which defaults to 'stable'. All PHP versions will be tested against the low + stable PHPCS version with stable PHPCSUtils.

A few extra builds are added to test against the missing combi of PHPCS stable + low PHPCSUtils. And then yet some more builds are explicitly added to test against dev versions of both tools.

This should make sure all relevant dependency combis are sufficiently tested and sets us up with more flexibility for the future.

Includes adding new (debug) step to display the output of the `composer info` command to make it more straight-forward to see exactly which versions of dependencies ended up being installed.

Notes:
* It could be considered to mark the "dev" dependency builds as "allowed to fail" (`continue-on-error`), but I've decided against that as the sniffs should always be ready for the next minor/patch release of the CS dependencies.

### DIFF
--- a/.github/workflows/basics.yml
+++ b/.github/workflows/basics.yml
@@ -39,10 +39,12 @@ jobs:
       - name: Validate Composer installation
         run: composer validate --no-check-all --strict
 
+      # Use the WIP/develop branches of the Extra CS dependencies as an early detection system for bugs upstream.
       - name: 'Composer: adjust dependencies'
-        run: |
-          # Using PHPCS `master` as an early detection system for bugs upstream.
-          composer require --no-update squizlabs/php_codesniffer:"dev-master" --no-interaction
+        run: >
+          composer require --no-update --no-scripts --no-interaction
+          squizlabs/php_codesniffer:"dev-master"
+          phpcsstandards/phpcsutils:"dev-develop"
 
       # Install dependencies and handle caching in one go.
       # @link https://github.com/marketplace/actions/install-php-dependencies-with-composer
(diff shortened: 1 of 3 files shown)


jrfnl:
Branch protection settings need updating to merge this PR, but there are more CI changes upcoming, so for now, I'm removing most of the "required" builds until those other changes have been pulled & merged. The "required" checks will be re-enabled before the next release.

PR #385, opened 2025-09-21
:sparkles: New `Universal.WhiteSpace.FirstClassCallableSpacing` sniff

jrfnl:
This sniff checks the whitespace around the ellipsis in a first class callable and can be used to address the following rule from PER-CS:

> **4.8 Function Callable References**
> A function or method may be referenced in a way that creates a closure out of it, by providing `...` in place of arguments.
>
> If so, the `...` MUST NOT include any whitespace before or after. That is, the correct format is `foo(...)`.

Ref: https://www.php-fig.org/per/coding-style/#48-function-callable-references

The desired "spacing" before/after the ellipsis is configurable and defaults to "no space".

Includes fixer.
Includes unit tests.
Includes documentation.


jrfnl:
Rebased without changes to run CI based on current version of the GHA scripts/PHPCS rules.

PR #398, opened 2025-10-14
:sparkles: New `Universal.PHP.RequireExitDieParentheses` sniff

jrfnl:
This sniff checks that calls to `exit`/`die` use parentheses, even if no argument is given and can be used to address the following rule from PER-CS:

> 4.7 Method and Function Calls
> ...
> The `exit()` and `die()` functions SHOULD always be called with parentheses even if no argument is given to clearly distinguish them from an access to a constant named `exit` or `die`.

Ref: https://www.php-fig.org/per/coding-style/#47-method-and-function-calls

Includes fixer.
Includes unit tests.
Includes documentation.


jrfnl:
Rebased without changes to run CI based on current version of the GHA scripts/PHPCS rules.

PR #406, opened 2025-11-06
:sparkles: New `Universal.Attributes.BracketSpacing` sniff

### LINKED ISSUE #386: [FR] Attributes: sniff to normalize spacing on the inside of attribute brackets
## Setting the scene

PHP 8.0 introduced support for attributes via the `#[...]` syntax. At this moment, neither PHPCS itself, nor PHPCSExtra contain any sniffs to handle the formatting of attributes.

The [PER Coding Standard](https://www.php-fig.org/per/coding-style/) from FIG, since PER 2.0, outlines a set of rules for attribute formatting to comply with, so using those rules as a starting point would allow for creating an initial set of sniffs to address attribute formatting.

## Proposed new sniff: `Universal.Attributes.AttributeBrackets`

To address these rules [from PER](https://www.php-fig.org/per/coding-style/#121-basics):
> Attribute names MUST immediately follow the opening attribute block indicator `#[` with no space.
>
> The closing attribute block indicator `]` MUST follow the last character of the attribute name or the closing `)` of its argument list, with no preceding space.

### Notes for the implementation

The name and categorization of this sniff still needs a good think - it could also be something like `Universal.WhiteSpace.AttributeBracketSpacing`.

Sniff should probably get the following configurable properties to give users some flexibility:
* `public $spacing = 0;`
* `public $ignoreNewLine = false;`

Suggested error codes:
* `SpaceAfterOpener`
* if `$spacing > 0`: `NoSpaceAfterOpener`
* `SpaceBeforeCloser`
* if `$spacing > 0`: `NoSpaceBeforeCloser`
* if `$ignoreNewLine === true`: BlanklineAtStart => 1 new line = okay, more than 1 is not okay
* if `$ignoreNewLine === true`: BlanklineAtEnd => dito

When `$ignoreNewLine` would be `false` and if it is deemed okay not to differentiate the errors too much (bundle `SpaceAfterOpener` and `NoSpaceAfterOpener` into one error code `SpaceAfterOpener`), the [`SpacesFixer` from PHPCSUtils](https://phpcsutils.com/phpdoc/classes/PHPCSUtils-Fixers-SpacesFixer.html) could make it pretty straight-forward to write this sniff.

## Describe the solution you'd like
A new sniff as outlined above.

The sniff should be able to flag and auto-fix the spacing on the inside of the attribute brackets.
```php
#[AttributeName] // <= Okay for `$spacing` = 0.
#[ AttributeName ] // <= Error for `$spacing` = 0 for both the attribute opener + the closer.

#[ AttributeName ] // <= Okay for `$spacing` = 1.
#[AttributeName] // <= Error for `$spacing` = 1 for both the attribute opener + the closer.
#[   AttributeName   ] // <= Error for `$spacing` = 1 for both the attribute opener + the closer.
```

Also see the examples outlined in the PER documents (rules + migration guide).

## Additional context (optional)

This ticket is part of a series of tickets related to PHP attributes and is the result of a detailed analysis of the rules as outlined in PER 2.0, as well as a critical look at what's still missing rule-wise.

- [ ] I intend to create a pull request to implement this feature.

jrfnl:
New sniff to enforce a fixed number of spaces on the inside of attribute block brackets. By default, the sniff expects `0` spaces, but this is configurable via a custom ruleset using the `spaces` property.
Also by default, new lines will not be allowed. This setting can also be toggled by changing the `ignoreNewlines` property in a custom ruleset.

Additionally, when `ignoreNewlines` is set to `true`, the sniff will also check there are no blank lines at the start or end of the attribute block.

Note: the `$spacing` and `$ignoreNewlines` property names are chosen to be in line with commonly used property names as used in PHPCS itself.

This sniff can be used with the default settings to address the following rule from PER-CS:

> 12.1 Basics
> ...
> Attribute names MUST immediately follow the opening attribute block indicator `#[` with no space.
>
> The closing attribute block indicator `]` MUST follow the last character of the attribute name or the closing `)` of its argument list, with no preceding space.

Ref: https://www.php-fig.org/per/coding-style/#121-basics

Includes fixers.
Includes metrics.
Includes unit tests.
Includes documentation.

Fixes #386

### DIFF
--- a/Universal/Docs/Attributes/BracketSpacingStandard.xml
+++ b/Universal/Docs/Attributes/BracketSpacingStandard.xml
@@ -1,0 +1,54 @@
+<?xml version="1.0"?>
+<documentation xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://phpcsstandards.github.io/PHPCSDevTools/phpcsdocs.xsd"
+    title="Attribute Bracket Spacing"
+    >
+    <standard>
+    <![CDATA[
+    Enforce a fixed number of spaces on the inside of attribute block brackets. By default, 0 (no) spaces are allowed.
+
+    Optionally allows for new lines.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: No whitespace on the inside of attribute block brackets.">
+        <![CDATA[
+<em>#[</em>MyAttribute()<em>]<em>
+class Foo {}
+        ]]>
+        </code>
+        <code title="Invalid: Whitespace on the inside of attribute block brackets.">
+        <![CDATA[
+<em>#[ </em>MyAttribute()<em>  ]<em>
+class Foo {}
+        ]]>
+        </code>
+    </code_comparison>
+
+    <standard>
+    <![CDATA[
+    When the option to allow a new line is turned on, the sniff will verify that there are no blank lines at the start/end of the attribute block.
+    ]]>
+    </standard>
+    <code_comparison>
+        <code title="Valid: No blank lines at the start/end of an attribute block.">
+        <![CDATA[
+#[<em>
+    </em>MyAttribute()<em>
+<em>]
+class Foo {}
+        ]]>
+        </code>
+        <code title="Invalid: Blank lines at the start/end of an attribute block.">
+        <![CDATA[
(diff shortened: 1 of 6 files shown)


jrfnl:
Rebased without changes to get a passing build against the new CI requirements.

PR #446, opened 2026-06-19
New `Universal.WhiteSpace.NamedArgumentSpacing` sniff

rodrigoprimo:
This sniff checks the whitespace around the colon in named arguments.

The desired "spacing" before and after the colon is configurable via the `spacingBefore` (defaults to no space) and `spacingAfter` (defaults to a single space) properties.

The default values follow [PER-CS 2.0](https://github.com/php-fig/per-coding-style/blob/2.0.0/spec.md#47-method-and-function-calls).

Fixes #440.

### Suggested changelog entry
* New `Universal.WhiteSpace.NamedArgumentSpacing` sniff to enforce consistent spacing around the colon for named arguments in function calls.


Universal/Sniffs/WhiteSpace/NamedArgumentSpacingSniff.php
```diff
@@ -0,0 +1,119 @@
...
+use PHP_CodeSniffer\Util\Tokens;
+use PHPCSUtils\Fixers\SpacesFixer;
+use PHPCSUtils\Utils\Parentheses;
+
+/**
+ * Checks the spacing around the colon for named arguments in function calls.
+ *
+ * By default, the sniff enforces the PER Coding Style rule that there must be no space
+ * between the argument name and the colon, and exactly one space between the colon and
+ * the argument value.
```
jrfnl: I'm not sure this needs to be documented in the class docblock. Having this information in the commit message for this sniff to justify the default values should be enough, I think.

Universal/Sniffs/WhiteSpace/NamedArgumentSpacingSniff.php
```diff
@@ -0,0 +1,119 @@
...
+    public function process(File $phpcsFile, $stackPtr)
+    {
+        // The colon is guaranteed to be the next non-empty token after the T_PARAM_NAME token.
+        $colon = $phpcsFile->findNext(Tokens::$emptyTokens, ($stackPtr + 1), null, true);
+
+        $parenthesisCloser = Parentheses::getLastCloser($phpcsFile, $stackPtr);
+        if ($parenthesisCloser === false) {
+            // Parse error/live coding: the named argument is not inside a closed set of parentheses.
+            return;
+        }
```
jrfnl, line 74: I wonder if the parentheses check should be the very first condition in the sniff ?

```suggestion
        $parenthesisCloser = Parentheses::getLastCloser($phpcsFile, $stackPtr);
        if ($parenthesisCloser === false) {
            // Parse error/live coding: the named argument is not inside function call parentheses.
            return;
        }
```

Universal/Sniffs/WhiteSpace/NamedArgumentSpacingSniff.php
```diff
@@ -0,0 +1,119 @@
...
+
+        // Find the start of the value being passed, which must sit inside the parentheses.
+        $valueStart = $phpcsFile->findNext(Tokens::$emptyTokens, ($colon + 1), $parenthesisCloser, true);
```
jrfnl: Is there any added value to using `$parenthesisCloser` for the "end token" in this statement ?

```suggestion
        $valueStart = $phpcsFile->findNext(Tokens::$emptyTokens, ($colon + 1), null, true);
```

And if there is, why only check against the parenthesis closer ? and not check against a comma (separating multiple arguments) ?

On that note, maybe there should also be a parse error test with a missing value when passing multiple arguments ?
```php
foo( nameA: , nameB: 10 );
```

Universal/Sniffs/WhiteSpace/NamedArgumentSpacingSniff.php
```diff
@@ -0,0 +1,119 @@
...
+        $spacingBefore = (int) $this->spacingBefore;
+        $spacingAfter  = (int) $this->spacingAfter;
+
+        // Check the spacing between the argument name and the colon.
+        SpacesFixer::checkAndFix(
+            $phpcsFile,
+            $stackPtr,
+            $colon,
+            $spacingBefore,
+            'Expected %s between the named argument name and the colon. Found: %s.',
```
jrfnl: Just an idea and would make it so the error message text more closely matches the phrasing used in PERCS.

```suggestion
            'Expected %s between the argument name and the colon for named arguments. Found: %s.',
```

Universal/Sniffs/WhiteSpace/NamedArgumentSpacingSniff.php
```diff
@@ -0,0 +1,119 @@
...
+            'Named argument: space before colon'
+        );
+
+        // Check the spacing between the colon and the argument value.
+        SpacesFixer::checkAndFix(
+            $phpcsFile,
+            $colon,
+            $valueStart,
+            $spacingAfter,
+            'Expected %s between the named argument colon and the value. Found: %s.',
```
jrfnl: Similar to above:

```suggestion
            'Expected %s between the colon and the argument value for named arguments. Found: %s.',
```

Universal/Tests/WhiteSpace/NamedArgumentSpacingUnitTest.1.inc
```diff
@@ -0,0 +1,40 @@
...
+// phpcs:set Universal.WhiteSpace.NamedArgumentSpacing spacingBefore 1
+foo(name : $value);
+foo(name: $value);
+foo(name  : $value);
+// phpcs:set Universal.WhiteSpace.NamedArgumentSpacing spacingBefore 0
+
+// phpcs:set Universal.WhiteSpace.NamedArgumentSpacing spacingAfter 0
+foo(name:$value);
+foo(name: $value);
+// phpcs:set Universal.WhiteSpace.NamedArgumentSpacing spacingAfter 1
```
jrfnl: Would having tests with `spacingAfter` and/or `spacingBefore` set to more than 1 have any added value ? Or is this basically already covered via the use of the PHPCSUtils `SpacesFixer` ?

Also: should there be a test where both property values are changed to another value then the default ?

Universal/Tests/WhiteSpace/NamedArgumentSpacingUnitTest.1.inc
```diff
@@ -0,0 +1,40 @@
...
+foo(name: $value);
+$obj->method(label: /* comment */ $value);
+new MyClass(
+    first: $a,
+    second: $b,
+);
+
+/*
+ * Invalid spacing.
+ */
```
jrfnl, line 18: I wonder if having some more tests with function calls with multiple arguments woud have value ? Now there's only one test covering that.
Just spitballing, but maybe a mixed multi-line multi-argument function call with comments interlaced to stress-test ? (as in a pretty untidy function call which doesn't comply with any other rules)

Also: should there maybe be a test with PHP 8.4+ `exit`/`die` as a function call with a named argument ?
And maybe a test with an anonymous class instantiation ?

jrfnl: Oh... just thinking.. maybe also vary the value a little more as well ? I mean, as things are, all tests have a `T_VARIABLE` as the value, while named arguments is particularly useful for, for instance, boolean toggles, where the meaning of the boolean value is unclear from the function call alone.

Universal/Sniffs/WhiteSpace/NamedArgumentSpacingSniff.php
rodrigoprimo: I don't have a strong opinion here. I went ahead and removed the second paragraph.

Universal/Sniffs/WhiteSpace/NamedArgumentSpacingSniff.php
rodrigoprimo: Good point, moved this check to be the first condition in the sniff. I'm not sure about changing the code comment, as this sniff is not only about named arguments inside function calls. To me, the original version is a bit better and more precise. Is there anything that I'm missing?

Universal/Sniffs/WhiteSpace/NamedArgumentSpacingSniff.php
rodrigoprimo: > Is there any added value to using $parenthesisCloser for the "end token" in this statement ?

My idea is to bail out if there is nothing between the colon and the parenthesis closer to avoid potential fixer conflicts with other parenthesis closer spacing sniffs like `PEAR.Functions.FunctionCallSignature`. 

> And if there is, why only check against the parenthesis closer ? and not check against a comma (separating multiple arguments) ?

I forgot to consider this. I have updated the sniff to bail out as well when there is no argument value, and the next non-empty token is a comma. Let me know what you think of this part of the code now.

> On that note, maybe there should also be a parse error test with a missing value when passing multiple arguments ?

Good idea, I added this test, but with two spaces between the colon and the comma to document that the sniff bails out when there is no argument value to avoid conflicts with other spacing sniffs.

What do you think of this approach?

Universal/Sniffs/WhiteSpace/NamedArgumentSpacingSniff.php
rodrigoprimo: I'm not sure, to be honest. To me, the original version reads a bit better. The "for named arguments" at the end sounds a bit worse.

Universal/Sniffs/WhiteSpace/NamedArgumentSpacingSniff.php
rodrigoprimo: See above.

Universal/Tests/WhiteSpace/NamedArgumentSpacingUnitTest.1.inc
rodrigoprimo: > I wonder if having some more tests with function calls with multiple arguments woud have value ?

That makes sense. I went ahead and added the test you suggested.

> Also: should there maybe be a test with PHP 8.4+ exit/die as a function call with a named argument ?
And maybe a test with an anonymous class instantiation ?

:+1: 

> Oh... just thinking.. maybe also vary the value a little more as well ?

:+1:

Universal/Tests/WhiteSpace/NamedArgumentSpacingUnitTest.1.inc
rodrigoprimo: > Would having tests with spacingAfter and/or spacingBefore set to more than 1 have any added value ? Or is this basically already covered via the use of the PHPCSUtils SpacesFixer ?

In my opinion, this is already covered via the tests in PHPCSUtils: https://github.com/PHPCSStandards/PHPCSUtils/tree/69b7d58f7284fa61e4aeeceeb6fae7d89d9f0c8a/Tests/Fixers/SpacesFixer

> Also: should there be a test where both property values are changed to another value then the default ?

Yes, I think so. I added it, and I opted to use different values for `spacingAfter` and `spacingBefore`, so I ended up addressing the point above as well.

Universal/Sniffs/WhiteSpace/NamedArgumentSpacingSniff.php
jrfnl: > To me, the original version is a bit better and more precise. Is there anything that I'm missing?

Named arguments are only supported within function calls, nowhere else.

Also see: https://wiki.php.net/rfc/named_params

Universal/Sniffs/WhiteSpace/NamedArgumentSpacingSniff.php
```diff
@@ -0,0 +1,118 @@
...
+            // Parse error/live coding: the named argument is not inside a closed set of parentheses.
+            return;
+        }
+
+        // The colon is guaranteed to be the next non-empty token after the T_PARAM_NAME token.
+        $colon = $phpcsFile->findNext(Tokens::$emptyTokens, ($stackPtr + 1), null, true);
+
+        // Find the start of the value being passed, which must sit inside the parentheses.
+        $afterColonNonEmpty = $phpcsFile->findNext(Tokens::$emptyTokens, ($colon + 1), $parenthesisCloser, true);
+        if ($afterColonNonEmpty === false || $tokens[$afterColonNonEmpty]['code'] === \T_COMMA) {
```
jrfnl:
```suggestion
        $afterColonNonEmpty = $phpcsFile->findNext(Tokens::$emptyTokens, ($colon + 1), null, true);
        if ($afterColonNonEmpty === $parenthesisCloser || $tokens[$afterColonNonEmpty]['code'] === \T_COMMA) {
```

Universal/Sniffs/WhiteSpace/NamedArgumentSpacingSniff.php
```diff
@@ -0,0 +1,118 @@
...
+            // Parse error or live coding: no named argument value. Bail out here to prevent fixer conflicts with comma
+            // or closing parenthesis spacing sniffs.
```
jrfnl: Not blocking. Not an important suggestion.

```suggestion
            // Parse error or live coding: missing argument value.
            // Bail out to prevent fixer conflicts with comma or closing parenthesis spacing sniffs.
```

Universal/Sniffs/WhiteSpace/NamedArgumentSpacingSniff.php
```diff
@@ -71,6 +71,8 @@ public function register()
      */
     public function process(File $phpcsFile, $stackPtr)
     {
+        $tokens = $phpcsFile->getTokens();
```
jrfnl: You may want to move this down ? `$tokens` is not used until line 87

Universal/Tests/WhiteSpace/NamedArgumentSpacingUnitTest.1.inc
```diff
@@ -38,3 +50,10 @@ foo(name  : $value);
 foo(name:$value);
 foo(name: $value);
 // phpcs:set Universal.WhiteSpace.NamedArgumentSpacing spacingAfter 1
```
jrfnl: This line is kind of redundant - you are changing this setting again two lines down.

Universal/Tests/WhiteSpace/NamedArgumentSpacingUnitTest.1.inc
```diff
@@ -0,0 +1,59 @@
...
+
+// Comments around the colon are reported, but not auto-fixed.
+foo(name  /* comment */ : $value);
+foo(name:  /* comment */ $value);
+
+// phpcs:set Universal.WhiteSpace.NamedArgumentSpacing spacingBefore 1
+foo(name : $value);
+foo(name: $value);
+foo(name  : $value);
+// phpcs:set Universal.WhiteSpace.NamedArgumentSpacing spacingBefore 0
```
jrfnl, line 57:
```suggestion
// Resetting to the default value.
// phpcs:set Universal.WhiteSpace.NamedArgumentSpacing spacingBefore 0
```

Universal/Tests/WhiteSpace/NamedArgumentSpacingUnitTest.1.inc
```diff
@@ -0,0 +1,59 @@
...
+// phpcs:set Universal.WhiteSpace.NamedArgumentSpacing spacingAfter 0
+foo(name:$value);
+foo(name: $value);
+// phpcs:set Universal.WhiteSpace.NamedArgumentSpacing spacingAfter 1
+
+// phpcs:set Universal.WhiteSpace.NamedArgumentSpacing spacingBefore 2
+// phpcs:set Universal.WhiteSpace.NamedArgumentSpacing spacingAfter 3
+foo(name  :   $value);
+foo(name:$value);
+// phpcs:set Universal.WhiteSpace.NamedArgumentSpacing spacingBefore 0
```
jrfnl, line 68:
```suggestion
// Resetting to the default value.
// phpcs:set Universal.WhiteSpace.NamedArgumentSpacing spacingBefore 0
```

Universal/Sniffs/WhiteSpace/NamedArgumentSpacingSniff.php
rodrigoprimo: Fair point. I was thinking about things like class instantiation and attribute parameters, but I agree that, at the end of the day, they are function calls.

jrfnl: Maybe it can just be simplified a little ?

```suggestion
        $parenthesisCloser = Parentheses::getLastCloser($phpcsFile, $stackPtr);
        if ($parenthesisCloser === false) {
            // Parse error/live coding: not inside a closed set of parentheses.
            return;
        }
```

Universal/Sniffs/WhiteSpace/NamedArgumentSpacingSniff.php
jrfnl: Maybe, again, we should just simplify it a little more ?
```suggestion
            'Expected %s between the argument name and the colon. Found: %s.',
```

Universal/Sniffs/WhiteSpace/NamedArgumentSpacingSniff.php
jrfnl: Similar to above:
```suggestion
            'Expected %s between the colon and the argument value. Found: %s.',
```

Universal/Tests/WhiteSpace/NamedArgumentSpacingUnitTest.1.inc
```diff
@@ -0,0 +1,59 @@
...
+    :
+        'value'
+)]
+class Baz {}
+foo(
+    // Comment.
+    first :$a,
+    second:$b, // Comment.
+    third  :  $c,
+);
```
jrfnl, line 37: Good effort, but we may be able to make this a bit more messy still.... ? Think of it this way: this sniff can be used without any other function call formatting sniffs being run, so we may need to deal with pretty messy code...

```suggestion
foo(
    // Comment.
    first :$a,
           second:$b /* Comment. */ , third  :  $c
    , fourth:
        1
                 + // phpcs:ignore Stnd.Cat.Sniff -- for reasons
$d
);
```

Maybe not even replace your test, but add this as an additional one ?

Not saying this code sample is the ultimate test, but you get my thinking hopefully by seeing this sample ?

Universal/Sniffs/WhiteSpace/NamedArgumentSpacingSniff.php
rodrigoprimo: Done

Universal/Tests/WhiteSpace/NamedArgumentSpacingUnitTest.1.inc
rodrigoprimo: Done

Universal/Tests/WhiteSpace/NamedArgumentSpacingUnitTest.1.inc
rodrigoprimo: I agree that this is useful. I added this new test.

